PR #10, opened 2018-08-17
Add contributing md

gassmoeller:
As discussed this is a proposal for a CONTRIBUTING.md document, as recommended by Github (https://help.github.com/articles/setting-guidelines-for-repository-contributors/). It should give users an idea about how they can contribute to Rayleigh (and encourage them to do so). We can discuss the individual parts of this file if you prefer a different wording or a different policy altogether, the current version is mostly taken from the geodynamics/aspect repository and might not be entirely applicable to Rayleigh. Let me know what you think about it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,129 @@
+# Contributing to Rayleigh
+Rayleigh is a community project that lives by the participation of its
+members — i.e., including you! It is our goal to build an inclusive
+and participatory community so we are happy that you are interested in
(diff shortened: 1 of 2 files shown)


ljhwang:
There is an incorrect link in CONTRIBUTING.md:  1.  Bug reports: We keep track of all open issues related to Rayleigh here.  "here" points to the wrong page.

gassmoeller:
Thanks for checking Lorraine, fixed it in #11.

PR #24, opened 2018-09-05
Add a Jenkinsfile for the tester setup

gassmoeller:
As discussed at our last zoom meeting this is a file that tells CIG's jenkins testserver (https://singleb.cse.ucdavis.edu/) how to build rayleigh, and what to do for testing. Currently, it just compiles the code and runs the benchmark example case. It does all of that inside a docker container (which is destroyed after the test run), and I had a few problems running Rayleigh with MPI inside the container, which is why it is currently running in serial. I will open separate issues about the MPI issues and the docker containers.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,82 @@
+#!groovy
+
+pipeline {
+  agent {
+    kubernetes {
+      //cloud 'kubernetes'
+      label 'mypod'
+      containerTemplate {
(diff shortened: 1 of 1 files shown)


gassmoeller:
You can now see the test results for this PR above the merge button. Currently this only ensures that it compiles, and that the benchmark case runs. Since I limited the maximum number of iterations to allow for faster testing the results might be off though.

PR #29, opened 2018-09-17
Preventing seg faults by only deallocating arrays if they have been allocated.

cianwilson:
More generally stop creating output if no fields are selected.


src/IO/Spherical_IO.F90
```diff
@@ -1514,7 +1514,7 @@ Subroutine Write_Equatorial_Slices(this_iter, simtime)
         !//////////////////////////////////////////////
         ! DeAllocation
         If (my_nth_owned .gt. 0) Then
-            DEALLOCATE( equslice_outputs)
+            if (allocated(equslice_outputs)) DEALLOCATE( equslice_outputs)
```
gassmoeller: Same here

cianwilson:
Thanks @gassmoeller, updated capitalization for this pull request but note that it's not consistent within this file (e.g. lines 378 vs 380 or even lines 417 vs 418) so this is probably worth bringing up when defining coding standards.

PR #34, opened 2018-09-17
New vorticity code!!!

feathern:
I think this is a good one.


gassmoeller:

PR #36, opened 2018-09-17
Added quantity code "valentin" in vorticity files

skoutnev:


gassmoeller:

PR #37, opened 2018-09-17
New named vorticity variable: Kyle

kyle-augustson:


gassmoeller:
There is another file in this pull request (a copy of `Diagnostics_Vorticity_Field.F90`), is this something you want to keep, or was this just a backup?

PR #40, opened 2018-09-17
Here is my s-component of vorticity.

hirokemono:
\Omega_s


aurnou:
Hiro, does this title imply that you mean the cylindrical sub-component of the helicity u_s . (curl(u))_s ?

gassmoeller:

PR #42, opened 2018-09-18
Fixing diagnostic document

sglane9891:
This commit fixes the tables in the diagnostic plotting documents. Auxiliary LaTeX files were deleted. Some typos were fixed.


gassmoeller:
I should add that this PR only contains a part of the original content, it now only updates the .gitignore file to exclude temporary and auxiliary files.

Issue #13, opened 2018-08-24
Bug in Diagnostics_Linear_Forces.F90


orvedahl:
In line 72, "buoy_work_pp" should be changed to "buoy_work_mm"

PR #43, opened 2018-09-18
fix Issue #13: a typo in diagnostic logic for buoyancy term.

orvedahl:
The logic previously asked if the prime-prime component was wanted, but the mean-mean is actually calculated. Now the logic asks if the mean-mean component is wanted.


gassmoeller:
I know too little about the code to determine if this is correct. This is something @feathern will need to check.

PR #50, opened 2018-09-18
Add latex styles for a more functional manual.

bangerth:
Specifically, use better defaults for the 'listings' and 'hyperref'
styles.

### DIFF
--- a/doc/src/user_guide/Rayleigh_User_Guide.tex
+++ b/doc/src/user_guide/Rayleigh_User_Guide.tex
@@ -1,20 +1,19 @@
 \documentclass[10pt, letterpaper]{article}
-\usepackage[letterpaper,margin=0.75in]{geometry}
 \usepackage{graphicx}
 \usepackage{amsmath}
 \usepackage[table]{xcolor}
 \usepackage{color}
-\usepackage[hidelinks]{hyperref}
 \setlength{\parindent}{0pt}
 \setlength{\parskip}{5pt}
 
+% use a larger page size; otherwise, it is difficult to have complete
+% code listings and output on a single page
+\usepackage{fullpage}
+
+
+% use the listings package for code snippets. define keywords for prm files
+% and for gnuplot
 \usepackage{listings}
-
-
-\definecolor{dkgreen}{rgb}{0,0.6,0}
-\definecolor{gray}{rgb}{0.5,0.5,0.5}
-\definecolor{mauve}{rgb}{0.58,0,0.82}
-
 \lstset{frame=tb,
   language=Fortran,
   aboveskip=3mm,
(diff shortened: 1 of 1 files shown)


ljhwang:
We have been working on a latex template for manuals.  This should appear in the geodynamics Best Practices repo shortly.  This is a near relative of the ASPECT manual so i suspect the style follows closely your proposal.

doc/src/user_guide/Rayleigh_User_Guide.tex
```diff
@@ -31,6 +30,33 @@
   breakatwhitespace=true,
   tabsize=3
 }
+\lstset{
+  language=C++,
```
gassmoeller: Is there a fortran option? Or does it not matter, which language is actually selected?

doc/src/user_guide/Rayleigh_User_Guide.tex
```diff
@@ -31,6 +30,33 @@
...
+  showstringspaces=false,
+  basicstyle=\small\ttfamily,
+  columns=fullflexible,
+  keepspaces=true,
+  frame=single,
+  breaklines=true,
+  postbreak=\raisebox{0ex}[0ex][0ex]{\hspace{5em}\ensuremath{\color{red}\hookrightarrow\space}}
+}
+\lstdefinelanguage{prmfile}{morekeywords={set,subsection,end},
+                            morecomment=[l]{\#},escapeinside={\%\%}{\%},}
```
gassmoeller: I do not think we need the prmfile language here. We could still keep it and modify it for the Rayleigh inputfile later. What is your opinion?

doc/src/user_guide/Rayleigh_User_Guide.tex
```diff
@@ -31,6 +30,33 @@
...
+\lstdefinelanguage{gnuplot}{morekeywords={plot,using,title,with,set,replot},
+                            morecomment=[l]{\#},}
+
+
+% use the hyperref package; set the base for relative links to
+% the top-level aspect directory so that we can link to
```
gassmoeller: aspect -> Rayleigh

doc/src/user_guide/Rayleigh_User_Guide.tex
```diff
@@ -31,6 +30,33 @@
...
+% files in the aspect tree without having to specify the
```
gassmoeller: aspect -> Rayleigh

doc/src/user_guide/Rayleigh_User_Guide.tex
bangerth: I don't think any is selected by default. This block just defines the C++ style -- which we don't need here, but I thought I'd add it as an example for how one would define another language style. The f90 style is already defined a few lines above.

This isn't important, so I removed the block.

doc/src/user_guide/Rayleigh_User_Guide.tex
bangerth: Yes, that was the idea.

bangerth:
All fixed.

PR #53, opened 2018-09-18
deleted auxiliary files and tex-file not used anymore

sglane9891:
As discussed.


gassmoeller:
I checked the remaining .tex files, and they are clearly leftovers, so this is good to go.

PR #59, opened 2018-09-18
Add a first automatic testcase

gassmoeller:
This is a first version of an automatic testcase. It is not scalable to more tests yet (I need to write a Makefile for that), and the only test is not very accurate, but it is better than the current version, that only checks if things run (not if they are the same as before).


Jenkinsfile
```diff
@@ -45,26 +45,19 @@ pipeline {
...
-          // also reduce runtime of the model for fast testing
           sh '''
-            sed \
-              --in-place \
-              -e 's/nprow = 2/nprow = 1/' \
-              -e 's/npcol = 2/npcol = 1/' \
-              -e 's/max_iterations = 40000/max_iterations = 400/' \
-              main_input
-          '''
+            cd tests/benchmark_diagnostics_input
```
tjesser-ucdavis-edu: Did you mean to get rid of the input_examples `cp` and `sed`?

gassmoeller: Yes, I added a new directory `tests/` that will collect a number of different subdirectories, which will each contain one particular testcase. The file I added for the first test is essentially the file we used before, including the sed commands. This just makes the Jenkinsfile less complex, and less specific to this particular testcase.

feathern:
Hi Rene,

I was about to approve this, but the NaNs in the accuracy summary caught my eye.   Is it possible to run this with nprow = 2 and npcol =2?    There is a bug in the code presently that causes faulty results when run in serial mode.   Two MPI ranks in each direction are needed.  This is something I hope to dig into someday, but in practice, running on four cores isn't a big deal.   Here's a copy of my accuracy output for reference if I run with 2+ cores in each direction:

``` ///////////////////////////////////////////////////////////////////////////
               RAYLEIGH ACCURACY BENCHMARK SUMMARY               
  
   Benchmark:  Christensen et al. 2001  (Non-MHD, Case 0)
  
   Radial Resolution      N_R =           48
   Angular Resolution N_theta =           64
  
   Averaging Interval (Viscous Diffusion Times) :       0.030000
  
   Beginning Iteration :          100
   Ending Iteration    :          400
   Number of Samples   :            4
 ----------------------------------------------------------------------------
   Observable      |    Measured    | Suggested   | % Difference |  Std. Dev.
 ----------------------------------------------------------------------------
   Kinetic Energy  :     28.015316     58.348000    -51.985816     13.571364
   Temperature     :      0.339570      0.428120    -20.683445      0.027416
   Vphi            :     -6.022263    -10.157100    -40.708837      2.167046
   Drift Frequency :     11.007091      0.182400   5934.589314      3.655992
```

feathern:
I should add that the NaNs concern me because if there is another bug, unrelated to the known serial bug, that produces NaNs, we might not catch it.
Otherwise, I think this is a very sensible first test case.  Thanks for working on this!

gassmoeller:
Ok, I rebased this to the current version. It sits on top of #128 so that should be merged first. But afterwards, this runs a simple test case on 4 cores that only takes 1 min and checks the output statistics. This will be the first accuracy regression test we have, so even if it can be improved in terms of setup and runtime I would like this to be merged as soon as possible (to make sure all other modifications dont break stuff).

gassmoeller:
Ok, I renamed the test case and the test succeeded. Ready to merge.

PR #60, opened 2018-09-18
Add initial version of a cmake configuration system.

bangerth:
This probably gets us more than 50% there. What's missing:
* Add the C source files. Currently, src/CMakeLists.txt only lists the F90
  source files.
* Deal with the fact that compilation fails for me right now. I'll copy
  the compiler error in a second.
* Whatever else comes up.

Either way, it's overall not a terrible amount of work or code.


bangerth:
This is the error message I get. I have no F90-fu, so don't know what all of this means.
```
Rayleigh/build> make -j8 -k
[  1%] Building Fortran object src/CMakeFiles/rayleigh.dbg.dir/Math_Layer/Legendre_Transforms_newlayout.F90.o
/home/bangerth/p/deal.II/1/projects/Rayleigh/src/Math_Layer/Legendre_Transforms_newlayout.F90:814:8:

     Do f = 1, nfield
        1
Error: Symbol ‘f’ at (1) has no IMPLICIT type
/home/bangerth/p/deal.II/1/projects/Rayleigh/src/Math_Layer/Legendre_Transforms_newlayout.F90:813:8:

     iend = nr
        1
Error: Symbol ‘iend’ at (1) has no IMPLICIT type
/home/bangerth/p/deal.II/1/projects/Rayleigh/src/Math_Layer/Legendre_Transforms_newlayout.F90:815:9:

     Do im =1, 2
         1
Error: Symbol ‘im’ at (1) has no IMPLICIT type
/home/bangerth/p/deal.II/1/projects/Rayleigh/src/Math_Layer/Legendre_Transforms_newlayout.F90:812:10:

     istart = 1
          1
Error: Symbol ‘istart’ at (1) has no IMPLICIT type
/home/bangerth/p/deal.II/1/projects/Rayleigh/src/Math_Layer/Legendre_Transforms_newlayout.F90:54:23:

   ylm(i,1) = (0.25d0/pi)**0.5d0 !00
                       1
Error: Symbol ‘pi’ at (1) has no IMPLICIT type
/home/bangerth/p/deal.II/1/projects/Rayleigh/src/Math_Layer/Legendre_Transforms_newlayout.F90:343:16:

  delta = secnds(t1)
                1
Error: ‘x’ argument of ‘secnds’ intrinsic at (1) must be of kind 4
/home/bangerth/p/deal.II/1/projects/Rayleigh/src/Math_Layer/Legendre_Transforms_newlayout.F90:350:16:

  delta = secnds(t1)
                1
Error: ‘x’ argument of ‘secnds’ intrinsic at (1) must be of kind 4
/home/bangerth/p/deal.II/1/projects/Rayleigh/src/Math_Layer/Legendre_Transforms_newlayout.F90:357:16:

  delta = secnds(t1)
                1
Error: ‘x’ argument of ‘secnds’ intrinsic at (1) must be of kind 4
/home/bangerth/p/deal.II/1/projects/Rayleigh/src/Math_Layer/Legendre_Transforms_newlayout.F90:364:16:

  delta = secnds(t1)
                1
Error: ‘x’ argument of ‘secnds’ intrinsic at (1) must be of kind 4
```

bangerth:
I suspect that the error is due to wrong code, not because of the cmake system. It may be worthwhile fixing this independently.

bangerth:
This works now -- the command sequence
```
  mkdir build
  cd build
  cmake ..
  make
```
yields an executable in `$BUILD/src` that executes correctly.

feathern:
Regarding the error in legendre_transforms_newlayout.f90,  that's a deprecated file that I will remove shortly in another commit.   It's something left over from the early development stages of Rayleigh.

feathern:
I'll have a look at this soon.  In general, I'm good with getting a Cmake build process working, but as someone who's never set up a cmake system, I'm uncomfortable with us requiring (just yet) that future commits are cmake compliant.  Can you or @gassmoeller give a quick summary of what would be required of a developer who adds a new source-code file to Rayleigh?

gassmoeller:
As far as I understand you would need to add a new line to `src/CMakeLists.txt` with the file path. In ASPECT we add all source files automatically, Wolfgang did you decide to not do that because of the error you described above?

bangerth:
I just went with the easy route because I didn't know what files are what -- this took some iterations, some files don't compile and some seem to be header files. The issue with globs for filenames is also that it requires you to re-run cmake if you add a file -- something that is easily forgotten and then you have the file, but it isn't compiled -- that's hard to understand. So I think listing files by hand is an easier approach.

But in general, @gassmoeller is right: All that's necessary to do is list the new file in the `CMakeLists.txt` file.

bangerth:
I tend to think that in the medium term, you probably want to get rid of the existing `Makefile` approach. It's not useful to maintain two methods in parallel.

PR #69, opened 2018-09-19
Additional Comments in Checkpointing.F90

feathern:
Added a few comments to checkpoints.F90 that describe what is being done during the checkpoint reading.

### DIFF
--- a/src/Physics/Checkpointing.F90
+++ b/src/Physics/Checkpointing.F90
@@ -512,9 +512,10 @@ Contains
                 my_in_disp = my_in_disp*nlm_total_old 
                 full_in_disp = nlm_total_old*n_r_old
             
-                !tnr is the number of radial points for this row x 2 (for complex values)
-                ! The rowstrip will hold all fields and AB terms for the RHS (hence factor 
-                !  of 2 below).
+                ! tnr is the number of radial points for this row x 2 (for complex values)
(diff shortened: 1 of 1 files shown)


src/Physics/Checkpointing.F90
```diff
@@ -509,7 +512,9 @@ Subroutine Read_Checkpoint(fields, abterms,iteration,read_pars)
                 my_in_disp = my_in_disp*nlm_total_old 
                 full_in_disp = nlm_total_old*n_r_old
             
-
+                !tnr is the number of radial points for this row x 2 (for complex values)
```
gassmoeller: Please add a space in the beginning to be consistent.

src/Physics/Checkpointing.F90
```diff
@@ -509,7 +512,9 @@ Subroutine Read_Checkpoint(fields, abterms,iteration,read_pars)
...
+                ! The rowstrip will hold all fields and AB terms for the RHS (hence factor 
+                !  of 2 below).
```
gassmoeller: Which factor of 2 are you referring to? The one you mentioned in the line before? But I thought that is because the values are complex. Do you mean RHS terms are always complex, and that is why you need the factor 2?

feathern:
I cleaned up the comment text.   The additional factor of two is due to the AB arrays.
-Nick

PR #72, opened 2018-09-19
move look up table calculation to a function that is used by all classes

orvedahl:
In the rayleigh_diagnostics.py file, each class calculated the look up table. This moves the calculation to a centralized function and does some basic error checking.


orvedahl:
You are correct; it is never used and I don't think it will be useful in the future. I'll remove it and commit shortly.

PR #74, opened 2018-09-19
fixed indentation of the source code

sglane9891:
I fixed the indentation in the source files. I replaced tab stops by four spaces and replaced tab stops at end of files. Two files required manual adjustments because the lines were too long.


sglane9891:
I removed the tab stops only in the source folder. I deleted the file src/Diagnostics/q.log which was another log-file written by pdflatex.

bangerth:
These are the real changes: https://github.com/geodynamics/Rayleigh/pull/74/files?w=1

gassmoeller:
Ah nice, I forgot about the whitespace setting. Good to go.

PR #79, opened 2018-09-21
write preamble and build information to a separate output file

orvedahl:
The preamble and some other build information is now output to a separate file at runtime. All the namelist options are also output into the file. With a small modification, this file can directly be used as the main_input for an identical run. The build information includes compiler information, git hash/branch information, custom directory locations, build date/time, build machine. The Makefiles get some build information and git information from the shell and insert it into the F90 file before it is compiled. I added a python utility into the src/Utility directory that ensures the lines do not get too long.

I do not have access to a Mac, and was not able to verify it works there.


src/Makefile
```diff
@@ -51,6 +51,10 @@ all:
 rayleigh_exec: $(OBJ)
 	$(FC) $(FFLAGS) -o  $(PROG) $(OBJ) $(LIB)
 
+# the lines that contain
+#    cp $(CUSTOMROOT)/*.F . || :
+# are basically logical OR statements, but the ":" shell character is always True.
+# So the end result is to suppress any and all errors from a failed cp command
```
gassmoeller: Please mention here why you did this. Is it necessary? What would happen if you would not do this?

src/Physics/Main.F90
```diff
@@ -59,9 +61,11 @@ Program Main!
         Call Test_Lib()
     Else
         Call Main_Initialization()
+        Call Set_Run_Parameters()
```
gassmoeller: Based on the name of the function when I read this I thought: Huh, did Rayleigh not read its input parameters if not in test mode? That sounds like a serious bug. Then I realized that this function actually *writes* the run parameters into a file, it does not *set* them for this run. Maybe `Write_Run_Parameters` would be a clearer name? What do you think?

orvedahl:
@gassmoeller I'm having trouble viewing your comments in github (I think I made them disappear somehow?), otherwise I would respond to them directly instead of a new comment. Here are my responses:

The function name --- yes, I agree it is misleading. The newest commit changed the name to Write_Run_Parameters

The comment in the Makefile --- The comment is a reminder of what the commands do. Those commands are included to protect against two situations; 1) the custom directory exists, but there are no F90 files in the directory. In this case, the build system would fail because the copy commands failed. 2) the custom directory does not exist. In this case, the copy command fails and so does the build process. Not including the extra shell commands forces the copy command to return a "success" status even if the files/directories do not exist.

We can definitely ask two questions: 1) should the build process go forward if the user gives a bad custom directory? 2) should the build process go forward if the given custom directory does not have any *.F90/F/c/py files in it?

gassmoeller:
I think you accidentally clicked on `resolve conversation` below each comment and that made them disappear. I reopened them. In general whenever you change the lines I wrote comments on, they will be hidden by github, because it assumes you addressed the comments I made. But you can always get them back by clicking on the `Show outdated` button to the right of the line where I wrote the comments.

Ok, I think in order to answer your question we should discuss the meaning of the CUSTOMROOT. Do you expect users to use that if they want to add their own files to Rayleigh? Or if they want to build Rayleigh in a separate folder? And would unexpected things happen if a user puts a typo into their CUSTOMROOT? If so, then the build process should probably fail, ideally with a helpful error message (e.g. `$CUSTOMROOT not found.` , or `$CUSTOMROOT does not contain expected files.`). Failing early is always preferable to providing unexpected results.

orvedahl:
Thanks, I will keep that in mind next time.

Nick's original idea was to provide users the ability to overwrite any of the source code files with their own versions, while still keeping the original Rayleigh files unmodified. The CUSTOMROOT capability allows users to store the modified files in a completely separate location, it has nothing to do with where the code is built.

When a user actively supplies a CUSTOMROOT directory, they expect the extra files to be included in the build process. As it currently stands, if there is a typo in the CUSTOMROOT directory, then the expected files will not be included, but the build will proceed. I would classify this as an unexpected result, as the user specifically included a CUSTOMROOT location. I think the build should fail if CUSTOMROOT does not exist. Furthermore, I think the build should fail if there are no files in the directory. I will add that in the next commit and be sure to include meaningful error messages.

feathern:
Correct me if I'm wrong, but Ryan's commit doesn't do anything with CUSTOMROOT (i.e., this isn't capability that he added; this was existing capability).  Instead, I think he just added a comment regarding some logic in the configure file.  Is this much true?

orvedahl:
Nick, you are correct, I'm just trying to make the new feature compatible with the existing CUSTOMROOT. I actually stole the copy command from the $(RayleighDir)/Makefile at line 30 or so. The new capability that I am adding has some version-specific files, debug vs optimized. So I need to copy files from CUSTOMROOT each time you build a different version in the $(RayleighDir)/src/Makefile.

The new capability is pretty much just the python parsing utility and the Run_Parameters.F90.

feathern:
Great, so I suggest that discussion of CUSTOMROOT is besides the fact here.    I have two suggested edits that I would like to see before we accept this, along with a longer-term question that doesn't need to be addressed right now:
1.) In the object_list file, please simply include Run_Parameters.o in MAIN_OBJ so that the semantics more or less match what's done for every other .o file (i.e., every other .o file is included in one of the predefined variables).
2.) Unless I'm missing something, write_run_parameters could be called from within the input module.  Is this possible?  If so, we should do this because writing the run parameters is really a side-effect of reading in the input.
3.)  This is the long-term question that doesn't need to be addressed right now, but I'm uncomfortable using python in the build process for what is otherwise a natively fortran 2003 code (soon to be F2008, and of course there's one C routine).   I think we should look into what the Fortran 2008 standard provides regarding compiler flags etc.  I seem to recall that F08 provides that information through some module, but of course the 08 standard isn't yet implemented completely on all compilers.

Anyway, I would say address (1) and if possible (2), and let's accept this commit.  Thanks for the hard work Ryan.
-Nick

orvedahl:
I moved the Run_Parameters.o file in the object_list and moved the call to write_run_parameters() to the Input module. The Main_Input subroutine now reads the namelists, checks for command line arguments, and then writes out the information.

As for point 3) above, I have two responses. 1) The sole purpose of the python routine (which does not import any fancy modules) is to take the long character lines and split them over multiple lines with proper Fortran syntax. This *should* be straight forward to implement in C, but I am much faster coding in python compared to C (even with a simple task such as this). 2) I am extracting some information using shell commands, such as the git commit, the git branch, and build date/time. It's unclear to me the best way to access this information from Fortran. This is all done inside the Makefile though, using simple bash commands (date, git, uname, pwd, which). So perhaps writing a C code in place of the python code solves this problem too.

gassmoeller:
Ah, I am happy to concede that I know little of this, so I am fine with this whenever Nick is happy. Did Ryan address your comments?
I have one more comment though, it looks as if `make install` does not longer work on the tester with these changes. Is there anything that broke it, or do we need to look into the test setup? See https://singleb.cse.ucdavis.edu/blue/organizations/jenkins/Rayleigh/detail/PR-79/10/pipeline

orvedahl:
@gassmoeller, the "make install" failed because the "make" failed. Part of the pull request involves running a very simple python routine from within the makefile. The build process in the tester is failing because it does not recognize the "python" command. I have been using this branch on my local machine and multiple supercomputers without any build issues.

feathern:
Does the testing container lack python entirely, or is python or is the issue that it is Ubuntu, which appears to use the alias python3 rather than python, since (the latest versions at least) seem to use python 3.x?   RHEL 7, by contrast appears to use python 2.x by default, and provides symlinks named python and python2.    

I appreciate the hard work here @orvedahl , but I'm wondering if there isn't some other means by which we can do this without python.  The iso_env module provides two functions which return a string containing the compile flags (though not libraries linked) and the compiler version.   I'm wondering maybe we can set values at compile time that map to preprocessor variables using -D switches.

orvedahl:
I will continue looking for a suitable solution. I'd have to think about it more in depth, but the -D switch idea sounds very promising (I never remember that capability).

PR #81, opened 2018-09-24
Added Axial Field Diagnostics

skoutnev:
1)Increased custom codes to have a range of 500 codes as discussed. 
2)Added a separate _codes.F file for 100 turbulentKE codes (as per Brad Hindman's approval) which now start at 2701.
3)Added axial field diagnostics which go from 2801-2814.
I tested by manually by looking at shell-slices fro each code after running the hydro benchmark (all of which looked reasonable). How did we decide to do rigorous testing?


orvedahl:
In lines 115, 126, and 135 of Diagnostics_Axial_Field.F90, I do not think the If statements are needed. The "Call Add_Quantity(...)" can happen at the same level as the DO_PSI loop, it does not need its own If statement.

The test that I used for the Kinetic Helicity components should be in the src/Diagnostics/Verification/ directory. I included the simple main_input file and the python code I used to do the test. Since these are all analytic quantities, you could output the velocity and derivative components, then construct v_z, vort_z, etc. manually. And compare your construction to the diagnostic output from Rayleigh (that is what my test code does).

feathern:
Hi Valentine,
I agree with Ryan regarding the conditionals.    I also think they're redundant.   Can you fix that, and then we can accept this commit.
-Nick

gassmoeller:
This PR has accumulated conflicts with the master version. Valentin could you take a look? Let me know if you need help with finding or fixing them.

feathern:
I just resolved the conflict.   I took the reference to the turbulent ke diagnostics indices file out.  @BWHindman is working on a commit that will populate that file properly soon.  For now, I've left the codes out in diagnostics_base.F90.

feathern:
Final comment here.   It seems as though the Jenkins check has passed, but it's stalled in reporting to github?  I'm a little unclear, but given that it's being run in the buggy serial mode as I mentioned in the comments of pull requst #59, I'm going to go ahead and merge this feature.

gassmoeller:
Merging was the right decision. Our Jenkins tester sometimes has a connection issue that we were not able to track down yet. Occasionally it fails to start (and report an error), and sometimes it does not report the result correctly (as in this PR). I and @tjesser-ucdavis-edu are still working on this and we will let you know once it is resolved.

PR #84, opened 2018-09-24
ensure configure only builds one executable with the -devel flag

orvedahl:
Modified configure to ensure that only a single executable is built with debugging flags. Previous behavior was to build both rayleigh.opt and rayleigh.dbg, but both would have debugging flags. This should work for both Intel and Gnu compilers.

### DIFF
--- a/configure
+++ b/configure
@@ -541,10 +541,12 @@ MULTIVEC=FALSE
 if [[ $DEVELMODE == "TRUE" ]]
 then
   echo "************************************************************************"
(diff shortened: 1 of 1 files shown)


gassmoeller:
Very nice. Did you test this with intel and gnu, or was that a it *should* work? :smirk:
I can test with gnu if you let me know.

orvedahl:
I originally tested it with only Gnu, but did not change anything within the Intel specific sections; hence the "it should work". However, I just verified that this works for both Intel and Gnu...only one debug version is compiled as expected.

PR #95, opened 2018-11-07
Revised the turbulent KE offset to be 500 past the custom diagnostics

feathern:
When accepting Valentin's axial diagnostics pull request, my conflict resolution inadvertently wiped out the 500 spaces he added to the custom diagnostics.   This minor change fixes that error.

PR #96, opened 2018-11-07
Included the Turbulent KE budget as a standard set of outputs, thereb…

BWHindman:
…y making them public and part of the standard output list.

1) Moves the code number definitions from Diagnostics_Base.F90 to turbKE_codes.F and added the tex description for each.

2) Added a few trivial outputs that were missing (shear production and azimuthal components of the flux)---involving changes to both turbKE_codes.F and Diagnostics_TurbKE_Budget.F90.

3) Removed some obsolete comment statements.

### DIFF
--- a/src/Diagnostics/Diagnostics_Base.F90
+++ b/src/Diagnostics/Diagnostics_Base.F90
@@ -91,27 +91,7 @@ Module Diagnostics_Base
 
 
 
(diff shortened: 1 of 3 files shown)


BWHindman:
I believe that this pull request makes Nick's recent fix #95 unnecessary.

PR #101, opened 2018-12-13
Sigterm

feathern:
I've added the capability for Rayleigh to intercept SIGTERM (kill -15) signals from the OS.  Many job schedulers will send a SIGTERM before sending SIGKILL (kill -9) to end the process, with some small amount of time (i.e. 30 seconds) in between.   This functionality allows Rayleigh to checkpoint and exit upon receiving a SIGTERM signal.

### DIFF
--- a/configure
+++ b/configure
@@ -32,12 +32,12 @@ function opt_flags {
   #$1 --version
   if [[ $1 == "INTEL" ]]
   then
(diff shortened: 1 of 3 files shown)


orvedahl:
This looks good to me. Nice addition.

PR #103, opened 2018-12-18
Reference Heating & Ohmic Heating

feathern:
This pull request involves three changes:
1)  The heating attribute of the reference state is now written to the reference file
2)  All attributes of the reference state object are initialized to zero at allocation time
3)  ohmic heating is now explicitly set to false in constant reference mode (Boussinesq).

Note also that I latexed the quantity code sheet to double check some quantities.  i did not update the one in the doc folder, however, because the modifications made to the latex tables back in September appear to have cut off several of the formulae.   This need to be addressed, but is outside the scope of this pull request.


feathern:
It's been three weeks now, with no review.  I am going to go ahead and merge this since the ohmic heating modification to Boussinesq mode is rather important.

gassmoeller:
Hi Nick, yes sorry for the non-reply. I was intimidated by the large number of line changes, and AGU and Christmas weren't helping my schedule. I now realize most of this change is actually documentation?

The rest looks ok (in particular the ohmic heating part), but I do not understand where the new ref%heating variable is used, except for writing it to the output. It does not seem to have any effect.

While the non-reaction is of course the responsibility of the reviewers, for future critical bug fixes it is helpful to put the critical part (in this case just one line?) in a separate pull request. That makes reviewing much faster.

feathern:
No problem.  Ryan was looking at these, but I think he's gotten busy.    Anyway, ref%heating is the reference heating (i.e., internal heating) function that is used in some models.  It's defined in ReferenceState.F90 and, if used, added to the RHS of the thermal energy equation in Sphere_Physical_Space.F90.
Cheers,
Nick

gassmoeller:
Makes sense.

PR #104, opened 2018-12-19
allow more AVX512 support

orvedahl:
Some AVX-512 systems require the MIC-512 flag, while some require the AVX-512, Rayleigh only used the AVX-512 flag before. Compiling with the AVX-512 flag and running on the MIC chips results in an error and Rayleigh never runs. This pull request compiles a 512-enabled binary that is optimized to run on both systems.

I have tested this extensively on the TACC Stampede computer that has both MIC-512 and AVX-512 chips; I used the exact same binary to get scaling results on both chip sets.

### DIFF
--- a/configure
+++ b/configure
@@ -574,7 +574,7 @@ then
         FFLAGS[0]=$OPT_FLAGS" -xCORE-AVX2"
         break            
        elif [ "$opt" = "AVX512" ]; then
-        FFLAGS[0]=$OPT_FLAGS" -xCORE-AVX512"
+        FFLAGS[0]=$OPT_FLAGS" -xCORE-AVX2 -axCORE-AVX512,MIC-AVX512"
         break      
        elif [ "$opt" = "ALL" ]; then
          DEFAULT='avx2'
@@ -583,7 +583,7 @@ then
          FFLAGS[0]=$OPT_FLAGS" -xCORE-AVX2"
          FFLAGS[2]=$OPT_FLAGS" -xAVX"
          FFLAGS[3]=$OPT_FLAGS" -xSSE4.2"
-         FFLAGS[4]=$OPT_FLAGS" -xCORE-AVX512"
+         FFLAGS[4]=$OPT_FLAGS" -xCORE-AVX2 -axCORE-AVX512,MIC-AVX512"
          RVALUES[0]=$DEFAULT 
          RVALUES[2]=avx
          RVALUES[3]=sse
--- a/configure
+++ b/configure
@@ -561,8 +561,9 @@ then
   echo " -- Sandy Bridge / Ivy Bridge cores :  AVX"
   echo " -- Haswell /Broadwell              :  AVX2 "
   echo " -- SkyLake / Knights Landing       :  AVX512 "
+  echo " -- Support for all of the above    :  FAT"
   echo ""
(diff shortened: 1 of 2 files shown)


feathern:
Hey Ryan,
This look reasonable to me, but I do have a quick question.  Why did you need to add the AVX2 flag as well?   That was unexpected.
-Nick

orvedahl:
I stole that directly from the TACC Stampede2 documentation:
https://portal.tacc.utexas.edu/user-guides/stampede2#building-for-performance-on-stampede2

It allows you to build on any Stampede2 node (login, KNL compute node, SKX compute node) and still build the AVX512 optimized binary. They say you could also do a more aggressive base option of "-xCOMMON-AVX512", while describing the "-xCORE-AVX2" option as reasonable though conservative.

For Stampede, the login nodes accept CORE-AVX512, so all nodes would be able to use the "-xCOMMON-AVX512" as the base instruction set. I do not know if that is the case for all supercomputers that have AVX512 chips. As a result, if you do the "-xCOMMON-AVX512" option, I'm not convinced that you would be able to build the AVX512 binary on a random login node without also including AVX2.

feathern:
Thanks for the link.  It confirms what I already thought.  Basically, you can compile for any target Intel architecture regardless of the architecture on which you are compiling by specifying the x-AVX flags.  Based on that link, it looks like TACC recommended AVX2 because they used to have Broadwell nodes, but AVX2 will also work on the Skylake nodes.

I would suggest removing the AVX2 flag, since it's redundant with the AVX2 that's already available via Rayleigh's configure process (this will not affect which nodes you can compile on).   It might also be useful for us to separately include a 'fat' option that compiles a binary will all avx instructions available.

-Nick

orvedahl:
TACC used to have Broadwell *login* nodes and replaced them with Skylake nodes. So including the AVX2 allowed you to run the executable on the Broadwell login node (for testing/installing I guess). I think TACC is special in that both the login node and the compute node accept AVX512, surely not every supercomputer will have a login node that accepts AVX512.

The "-x" option must be compatible with all the choices in the "-ax" list. I have not found a definitive answer if the architecture on which you are compiling matters or not. Can you compile a AVX512 binary on a AVX2 chip without including any AVX2 flags?

feathern:
Yes, you can compile AVX512 on an AVX2 chip.   I am positive about this because I did it routinely on CU's Summit machine.  This is the reason Intel included these flags (in place of just -xhost).  I think you are correct that the probable reason for TACC suggesting avx2 is so that applications whose build process involves small tests can have the tests pass on the login node.  I've run into situations on Summit where compiling a fat binary was necessary to avoid failed tests.   Rayleigh, however, doesn't execute any tests during its compile.  If we ever do, then we will need to consider modifying the configuration process.

orvedahl:
Perfect. I guess that makes a lot of sense, otherwise Intel would always suggest -xhost...

I'll replace the -xCORE-AVX2 with -xCOMMON-AVX512 and look into adding the "fat" option.

orvedahl:
Just tested the AVX512 and FAT options on TACC, both Skylake and Knights Landing successfully ran using the same binary.

For reference, choosing the FAT option took about 14 minutes to compile both the debug version and the optimized version (usually it's about 5 min), so be patient.

PR #106, opened 2019-01-08
Command-line Nondimensional Parameters

feathern:
This commit allows Ra, Pm, Pr, and E to be specified at the command line.  Calling syntax works as with other command line parameters.   Parameters are specified by their abbreviation, namely:

./rayleigh -Pr N -Pm M -Ra X -E Y

Note that due to my branching/merging practices, this pull request contains changes associated with my earlier run-timing commit.  That commit should be accepted first.


feathern:
Command-line parameters have always taken precedence over those set in the input file.   I believe this is mentioned in the documentation with regard to setting nprow, npcol, nr, and ntheta, but I haven't updated the documentation to include these.   The documentation is something that's come back to the forefront of my mind, and I will look into updating in sometime next month most likely.
-Nick

gassmoeller:
Ok, just checking if there is a consistent policy for command-line vs input parameters. But if that is what is done for the other parameters already, this is fine.

PR #107, opened 2019-01-10
Bug Fix:  Record numbering following restart

feathern:
This commit involves one bug fix and one small addition:

Bug fix:  There was a bug that was mentioned during the September hackathon.  Output files which were partially complete prior to restart were subsequently corrupted due to issues with the record numbering.   This commit fixes this bug.

Addition:  Normally I would make this a separate pull request, but I added this functionality while debugging the record-number issue.   Rayleigh now accepts the command-line flags -init and -mag_init which can be followed with the desired value of init_type and magnetic_init_type respectively (overriding the value set in main_input).  This can be useful for scripting in addition to debugging.  Calling syntax:  ./rayleigh -init_type -1  (to restart from a checkpoint, for example)

### DIFF
--- a/src/IO/Spherical_IO.F90
+++ b/src/IO/Spherical_IO.F90
@@ -3565,8 +3565,11 @@ Contains
 
             Call self%update_position                                ! save current position
             Read(self%file_unit,POS = 9)self%current_rec             ! read previous record #
+#ifdef INTEL_COMPILER 
+            fstat=fseek(self%file_unit, self%file_position, 0) ! return to end of file 
+#else
             Call fseek(self%file_unit, self%file_position, 0, fstat) ! return to end of file 
-
+#endif
 
             self%current_rec = self%current_rec+1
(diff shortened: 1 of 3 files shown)


src/IO/Spherical_IO.F90
```diff
@@ -3710,7 +3716,8 @@ End Subroutine CloseFile_Par
     Subroutine Update_Position(self)
         Implicit None
         Class(DiagnosticInfo) :: self
-        INQUIRE(UNIT=self%file_unit, POS=self%file_position)
+        !INQUIRE(UNIT=self%file_unit, POS=self%file_position)
```
gassmoeller: Is this line still necessary? Undocumented commented source code is always confusing to readers.

feathern: Oh, no - thanks.  I meant to delete that.   I expect to make another small pull request later this week.  I will delete that line when I do.
-Nick

PR #108, opened 2019-01-11
Read_CMD_Line

feathern:
Having added a few command-line parameters over the past week, I noticed that the semantics of doing so was cumbersome.  I've added a new interface named read_cmd_line which allows any new command-line parameters to be added by creating a single line of new code in Input.F90.   I've tested the functionality of the interface, but it might good for a second set of eyes to make sure I didn't misspell any of the preexisting parameter names.

-Nick

### DIFF
--- a/dev_build
+++ b/dev_build
@@ -1,4 +1,4 @@
 make distclean
-./configure -devel $1
+./configure -devel
 make
 make install
--- a/dev_build
+++ b/dev_build
@@ -1,4 +1,4 @@
 make distclean
-./configure -devel
+./configure -devel $1
 make
 make install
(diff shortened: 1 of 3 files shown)


orvedahl:
In general, this looks good to me, but I don't think the "-ntheta" flag is behaving the way it should (everything else seems to work). My main_input file specified l_max, then I tried to change the resolution with the "-ntheta" command line option without succeeding. If the main_input specifies n_theta, it works as expected.

feathern:
Hey Ryan,
The behavior you described is actually expected and isn't related to the command-line flags.   I had to choose whether l_max or n_theta took precedence if both were values set; for better or worse, I went with the former.   You can see this in Problemsize.F90 beginning at line 253.   As long as l_max is set (whether in main_input or at the command line), it will be used to determine n_theta.

Also, I just added a small update to this pull request.  l_max and aspect_ratio can now be set at the command line via -lmax and -aspect flags.

-Nick

orvedahl:
I knew l_max took precedence and figured that was what I was seeing, but wasn't sure that was the intended behavior given that command line options take precedence over namelist options. Makes sense to me.

PR #110, opened 2019-01-17
Benchmark Mode / Jobinfo.txt Bug Fix

feathern:
This is a minor bug fix.  I noticed last week that the jobinfo.txt file is created before benchmark-mode overrides are enforced.  When benchmark mode was on, the jobinfo.txt would display incorrect values for Rayleigh_Number etc.   I pulled the write_run_parameters() routine into Main.F90 to fix this.   I also made a few small style edits and removed a commented line in Spherical_IO.F90.

### DIFF
--- a/src/IO/Spherical_IO.F90
+++ b/src/IO/Spherical_IO.F90
@@ -3716,7 +3716,6 @@ Contains
     Subroutine Update_Position(self)
         Implicit None
         Class(DiagnosticInfo) :: self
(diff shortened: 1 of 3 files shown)


feathern:
Are you able to merge, Ryan?

orvedahl:
Oddly enough, it depends on what computer I use. For whatever reason the version of Firefox I have on my work station does not let me merge; the button is green and shows up, but nothing happens when I click it. But it looks like my laptop version of Firefox works just fine (I happen to have it today).

PR #113, opened 2019-01-31
Fix bug in ReferenceState that shifted all arrays by one element if r…

illorenzo7:
…eading an older "reference" file

I messed up in my previous fix of ReferenceState; I forgot about the two ints in the beginning of the binary 'reference' file, namely sig=314 and nr. So when I closed and reopened the file when handling an exception, the float arrays started in the wrong spot, meaning each ref.radius, ref.density, etc., had their first element from the array previous. This should be fixed by this commit, which reads in two ints after the 'reference' file is opened again. 

Sorry about that!


feathern:
Just saw this -- will look this weekend.

PR #119, opened 2019-02-12
Fixing diagnostic's documentation

sebglane:
This fixes the formatting problems in the diagnostic's documentation. I have adjusted the formatting of tables manually. This comes with a lot of changes of the underlying files. So a comparison of the differences in each file does not make much sense. I have merged files if possible and therefore some files were deleted.


feathern:
Thanks so much, Sebastian!  I will try to have a look at this on Thursday or Friday. -Nick

feathern:
Hi Sebastian,
Thanks again for this.  I just had a chance to look, and a recompile of the pdf seems to fix things.   A quick questions:  
1)  Did you modify the quantity-code tex files by hand?  Those might be overwritten the next time I run the Python routines that automatically generate the tables and latex math formulae.
2)  Is the real essence of the change that you use the single_column header file now, instead of the old header file?

Thanks!
-Nick

sebglane:
Hi Nick,

1. Yes I modified the ordering of the quantity-codes by hand to adjust them to the formatting of the tables. If you overwrite them the formatting will probably be broken. Can manually add new quantities to the tables without using the python script? Does the python script change the quantity codes?
2. There are two header files now. One for single column tables and one for two columns. For the quantities having long expressions a single column format is used. Otherwise tables with two columns are applied.
Sebastian

feathern:
Hi Sebastion,
Sorry to take so long to respond to this.   The overwriting of the tables (generated via a Python script) fixed this issue.  In the meantime, we've now moved to Sphinx documentation, which makes the tables much more legible and web friendly than they were before!
-Nick

PR #124, opened 2019-05-21
Added Lx and Ly quantity codes to Diagnostics_Angular_Momentum; and s…

illorenzo7:
…eparate quantity codes for L due to the convection (m>0) and mean flows (m=0). Also modified Problem_Size to produce the arrays phivals, cosphi, and sinphi.

More notes for this pull request: when the spherical-shell convection community has talked about "angular momentum leaks," we are usually concerned with L_z. However, we should probably also be concerned with L_x and L_y! Also, it may be helpful to see where the possible leakage is entering the system--either into the mean flows or the convection. I've thus added 8 new quantities to Diagnostics_Angular_Momentum. Nick added amom_z, and I have added amom_x, amom_y, and the corresponding codes for amomp (fluctuating flows) and amomm (mean flows). 

Here's to hoping they all stay zero!

### DIFF
--- a/src/Diagnostics/Diagnostics_Angular_Momentum.F90
+++ b/src/Diagnostics/Diagnostics_Angular_Momentum.F90
@@ -58,8 +58,9 @@ Contains
         If (compute_quantity(amom_x)) Then
 
             DO_PSI
-                qty(PSI) = ref%density(r)*radius(r)*(-costheta(t) &
-                    & *buffer(PSI,vphi) - sinphi(k)*buffer(PSI,vtheta))
+                qty(PSI) = ref%density(r)*radius(r)*&
+                    & (-costheta(t)*cosphi(k)*buffer(PSI,vphi) -&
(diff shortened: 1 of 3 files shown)


feathern:
Hi Loren,

Two comments:

1.)  I disagree with your formulae for Lx and Ly.   I think that the vphi term in Lx should be multiplied by cosphi(k)  and that the vphi term n Ly should be multiplied by sinphi(k).

2.)  Can you change the compacted loop notation in Problemsize.F90 to the standard (albeit multi-line) loop notation so that it agrees with the style used in the rest of the code?

-Nick

illorenzo7:
@feathern, thanks for the catch, somehow my cosphi/sinphi factors multiplying the v_phi term got lost. I put them in. I also changed the initialization of "phivals" to occur in a do-loop, as well as cosphi/sinphi. I *think* this is the style of the rest of the code now?

PR #125, opened 2019-05-22
Added directory python_util, enclosing routine write_reference write …

illorenzo7:
…custom reference profiles in binary such that Rayleigh likes reading them; Modified Custom_Reference to properly read in custom reference profiles, and removed the rayleigh_functions/rayleigh_constants framework from both ReferenceState and TransportCoefficients

MORE NOTES: Originally it looked like the Custom_Reference routine was geared toward reading variables into the arrays rayleigh_functions and rayleigh_constants, but this wasn't explicitly done in the code. Also TransportCoefficients disagreed on the dimensions for rayleigh_functions, since it expected the reference profiles AND the transport profiles to be there, while ReferenceState only allocated for the reference profiles. 

I have things set up now so that Custom_Reference properly reads in the reference state (not the transport coefficients) from the file "custom_reference_file" (specified in the Reference_Namelist of main_input) and interpolates and/or reverses the grid in this file, so that users can supply reference states on a linearly spaced, super-fine grid (I have been using nr_ref = 5000, and increasing radius) and Rayleigh will interpolate. 

I propose that if we also desire capability to read in custom transport coefficients, this is done in a separate routine Custom_Transport in TransportCoefficients, and specified in the Transport_Namelist of main_input. 

Finally, I created a Python routine write_reference so that users can easily convert the custom reference states they generate into a binary form that Rayleigh can understand. It didn't seem to belong in "post_processing" so I created a new directory "python_utils," which I am envisioning will eventually more Python routines that aren't associated with post-processing (e.g., Chebyshev derivatives/integrals, interpolation schemes, etc.) Happy to modify if this new directory is undesirable.

### DIFF
--- a/python_util/write_reference.py
+++ b/python_util/write_reference.py
@@ -1,0 +1,63 @@
+#
+#  Copyright (C) 2018 by the authors of the RAYLEIGH code.
+#
+#  This file is part of RAYLEIGH.
+#
+#  RAYLEIGH is free software; you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation; either version 3, or (at your option)
+#  any later version.
+#
+#  RAYLEIGH is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


illorenzo7:
Closing this because I didn't realize Nick and Lydia were working on something similar just recently!

PR #127, opened 2019-06-11
turbke_offset is now referenced properly to meq_offset.

feathern:
This adjustment references the turbulent kinetic energy codes to the magnetic energy codes.  Between these two code sets is the custom diagnostic block.  This was causing issues in the auto generation of quantity code tables (anything after the custom code block would fail to compile).


feathern:
Thanks, and yeah no commit should ever change existing codes again.  All new codes should fall within the existing free space or be appended to the end.

PR #128, opened 2019-07-18
Fix parallel test error

gassmoeller:
It looks like the problems we had with using OpenMPI inside the docker container that we use for testing do not occur with MPICH (at least locally when I run the container on my laptop). I would like to see if that is also the case on the jenkins server. If this works then we can start running some more test models for every pull request.

Please do not merge this until I have verified that it is ready.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,8 +2,8 @@
 
 pipeline {
   agent {
-    docker {
-      image 'geodynamics/rayleigh-buildenv-bionic:latest'
+    dockerfile {
+      dir 'docker/rayleigh-buildenv-bionic'
     }
   }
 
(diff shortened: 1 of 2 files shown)


gassmoeller:
Alright it looks like this works. To summarize what this PR does:
- We now build the docker container for testing directly from the Dockerfile in docker/rayleigh-buildenv-bionic, this saves the detour of somebody building an image from that dockerfile and uploading it to dockerhub to then download again on the tester
- The dockerfile now uses MPICH instead of OpenMPI. The version of OpenMPI that ships with Ubuntu 16.04 has caused us problems with other docker test setups before (they do not like running as root, what Jenkins requires, and they have problems with long filenames that happen when mapping directories into folders)
- Using MPICH instead of OpenMPI got rid of the error message that I only saw when running Rayleigh in parallel on the tester. It is unclear if that error was a real error or a problem with docker+openmpi, but since it was pretty obvious, and not reproducible outside the tester, I am pretty sure it was a false alarm.
- I now run the example model with 4 cores instead of 1.

When this is merged I can rebase #59, and we can start building a bigger testsuite during the hackathon.

PR #129, opened 2019-07-22
This fixes a typo

gassmoeller:
@feathern: You should take a look at this.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -41,7 +41,7 @@ Rayleigh requires the the following compilers and libraries:
 
     OR
 
-3a. BLAS
+3a. BLAS.46
 3b. LAPack
 3c. FFTW (version 3.x)
 


nicholasjnelson:
This is terrible.

INSTALL
```diff
@@ -41,7 +41,7 @@ Rayleigh requires the the following compilers and libraries:
 
     OR
 
-3a. BLAS
+3a. BLAS.46
```
cianwilson: This looks like a typo.

feathern:
Great!  But I think we're going another direction with these typo issues!

tukss:
Shouldn't this be `.42`?

gassmoeller:
> This is terrible.

Please keep comments friendly and constructive. E.g. write `please replace this by ...` or `In my mind this should look like this: ...`. This is a public repository and 26 users are currently following the discussions in this repository.

Comment to readers: We were having a in-person discussion about pull request review at the time of this PR, so all comments were friendly banter.

PR #130, opened 2019-07-22
Move sphinx documentation

gassmoeller:
I moved over the content of feathern/rayleigh_doc into the main repository. The only conflicting file was the Makefile and I kept a copy of the old file as `Makefile.manual`. The only functional change is the replacement of the directory path in `build_menu_pdf.py`. Surprisingly, everything seems to be working out of the box on my machine (html and pdf versions of the sphinx build compile and look correct). 

I would want to start with this PR without making large content changes, so we can set up the structure for everyone to improve the documentation later.


feathern:
Great -- thanks Rene!

PR #131, opened 2019-07-22
An attempt at a "generic" input implementation

cianwilson:
This pull request is not ready to be merged yet but I'm proposing it for discussion purposes.

It implements some python that users can interact with either as a module or as a script to describe an initial condition (and hopefully eventually a boundary condition, possibly a reference state?).  This python then generates a generic input file that can be read in by rayleigh in parallel using some fortran also implemented here.

At the moment this is only hooked up to set the temperature initial condition, in which case a user should specify:
```
&Initial_Conditions_Namelist
init_type = 8 
t_init_file = '<filename>'
/
```
where `<filename>` is the name of the file generated by the python script.

#### Generating the file:
For example, if a user wanted to specify a single mode in that input file then they could just run the script:
```
rayleigh_spectral_input.py -m 0 0 0 1.+0.j -o example
```
to specify (n,l,m) = (0,0,0) to have a coefficient 1.+0.j and output it to the file `example`.

This could also be done using the python as a module.  In a python shell this would look like:
```
from rayleigh_spectral_input import *
si = SpectralInput()
si.add_mode(1., n=0, l=0, m=0)
si.write('example')
```

For a more complicated example, e.g. the hydrodynamic benchmark from Christensen et al. 2001, the user can specify functions of `theta`, `phi` and `radius` that the python will convert to spectral:
```
rayleigh_spectral_input.py -ar 0.35 -sd 1.0 -nt 96 -nr 64 -o example \
    -e 'import numpy as np; x = 2*radius - rmin - rmax; 
rmax*rmin/radius - rmin + 210*0.1*(1 - 3*x*x + 3*(x**4) - x**6)*(np.sin(theta)**4)*np.cos(4*phi)/np.sqrt(17920*np.pi)'
```
in "script" mode.

Alternatively, in "module" mode in a python shell:
```
from rayleigh_spectral_input import *
si = SpectralInput(n_theta=96, n_r=64)
rmin, rmax = radial_extents(aspect_ratio=0.35, shell_depth=1.0)
def func(theta, phi, radius):
  x = 2*radius - rmin - rmax
  return rmax*rmin/radius - rmin + 210*0.1*(1 - 3*x*x + 3*(x**4) - x**6)*(np.sin(theta)**4)*np.cos(4*phi)/np.sqrt(17920*np.pi)
si.transform_from_rtp_function(func, aspect_ratio=0.35, shell_depth=1.0)
si.write('example')
```

Note that these two examples will have produced different data formats - the first one sparse (listing only the mode specified) and the second one dense (listing all modes).

NOTES:
 - python only tested with python3 so far


cianwilson:
Now python2 safe.

cianwilson:
@feathern: this should be good to review now.

We've expanded the initial conditions to all "primary" fields - T, W, Z, P, C & A.  (One question, does it make sense to have an initial condition on P??)

We've added a test that tests T, C & A using the Christensen et al. 2001 benchmark.  It runs the hard coded initial conditions (init_type and magnetic_init_type 1) for one timestep then compares that output to init_type and magnetic_init_type 8 with the relevant generic input files generated on the fly.

We've also added some documentation that discusses how to use the new functionality.

Thanks!

feathern:
Just noting that I'm not totally sure how much sense it makes to initialize P, BUT I think it's fine to have the functionality there in case someone  has a good reason to do so.

cianwilson:
Thanks @feathern!

PR #134, opened 2019-07-23
adapt makefiles to fix parallel make

tukss:
This introduces a dependency on the `makedepf90` tool to automatically
create the module dependencies.

`Legendre_Transforms_newlayout.F90` was deleted because it also provides
the `Legendre_Transforms` module leading to build errors.


tukss:
I deleted `Legendre_Transforms_newlayout.F90` to fix build errors. It seems that it was not used before, but someone more knowledgeable should comment.

tukss:
This introduces `makedepf90` as a dependency. This is a small tool written in C to generate the F90 dependency tree. If we don't want this as a dependency we could instead just commit the file `Makefile.fdeps` and update it manually every time the module structure changes. A third option might be to do this automatically with a hook.

gassmoeller:
Nice work, I like it. I tested this PR and it works for me. However, I think adding a new dependency is something we might not want, and @feathern should decide about. Another option is to check in the dependency file, and let the tester check for changes with every PR and report back when we need to update the dependency file.

tukss:
> Another option is to check in the dependency file, and let the tester check for changes with every PR and report back when we need to update the dependency file.

Yes, that sounds like a good compromise. The dependency file should not change that often anyway. We could add optional make target to recreate the dependency file.

tukss:
> I deleted `Legendre_Transforms_newlayout.F90` to fix build errors. It seems that it was not used before, but someone more knowledgeable should comment.

@feathern can you quickly comment on whether we need to keep the `Legendre_Transforms_newlayout.F90` file?

src/Makefile
```diff
@@ -7,11 +7,11 @@ include object_list  #List of object files and associated variables
 
 ######################################################
 
-MAKENP  = $(MAKE) --no-print-directory
+MAKEFLAGSS  += --no-print-directory
```
gassmoeller: Is it `MAKEFLAGS` instead of `MAKEFLAGSS`? I have to little experience with make to know if this is correct.

tukss: Good point. That is a typo. Actually `--no-print-directory` is already passed down from the top `make` invocation, so I'm just removing that line.

PR #136, opened 2019-07-23
Changes directory structure for User documentation for Sphinx.

ljhwang:
This creates a new directory for the User Manual for Sphinx. Minor fixes have been made to the files but errors still exist in dead links, formatting of equations that run over the edge, and equations are not numbered. 

In addition, table formatting should be checked in the conversion to pdf.


gassmoeller:
Thanks for the cleanup! Looks good. Ready to be merged when the tester has finished.

PR #140, opened 2019-07-23
Prettify pdf and html documentation

gassmoeller:
This PR fixes some style issues with the pdf documentation and removes some now not longer needed sections without contents.


gassmoeller:
Thanks!

PR #142, opened 2019-07-23
Update .gitignore and remove accidentally added files

gassmoeller:
As discussed with @tukss it is not really useful to keep jupyter checkpoint files around. They were likely accidentally added. Also prevent accidental additions of MacOS files.


gassmoeller:
:tada:

PR #143, opened 2019-07-23
improve nbstripout integration

tukss:
This fixes a few error messages and includes `nbstripout` into the CI Docker container.

### DIFF
--- a/etc/check_ipynb_cleared
+++ b/etc/check_ipynb_cleared
@@ -3,7 +3,7 @@
 if ! type nbstripout > /dev/null
 then
         echo "You need to install the nbstripout command."
-        echo "Use can the Python package manager of your choice (pip, conda, ...)."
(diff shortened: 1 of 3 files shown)


etc/check_ipynb_cleared
```diff
@@ -1,10 +1,18 @@
 #!/bin/sh
 
+if ! type nbstripout > /dev/null
+then
+        echo "You need to install the nbstripout command."
+        echo "Use can the Python package manager of your choice (pip, conda, ...)."
```
gassmoeller:
```suggestion
        echo "Use the Python package manager of your choice (pip, conda, ...)."
```

tukss:
> Would it be much work to create a similar check for the `make clear_ipynb` target? Maybe the target could just call the script instead of calling nbstripout directly?

In the new version we should always have the error message. I implemented clearing via a command line option.

PR #149, opened 2019-07-23
Automatically check python notebooks for output

gassmoeller:
This PR activates an additional check on the tester to automatically fail if one of the python notebooks in the repository contains output cells with output. This output can be large, is reproducible, and should therefore not be included in the repository.


gassmoeller:
:+1:

PR #154, opened 2019-07-24
Remove latex documentation.

gassmoeller:
This PR removes all of the old latex documentation. All of this should now be available on https://rayleigh-documentation.readthedocs.io/en/latest/ . 

@feathern should take a look before we merge this.


ljhwang:
Will we keep the rayleigh_doc repo around for awhile or will previous versions serve as an "archive"?

gassmoeller:
I think the `rayleigh_doc` repo only contains the diagnostic codes that are now also in this repo, therefore we do not need to keep it. The question what happens with the old manual is interesting. I would say having it in the old versions of this repository is enough of an archive. Do others have opinions?

PR #155, opened 2019-07-24
Cref example

nicholasjnelson:
Modified a version of Lydia's Jupyter notebook for creating a custom reference state so that is now loads a MESA profile and creates a custom reference state using the MESA model. In this case, the MESA model is a 1 M_sun zero-age main sequence star.


gassmoeller:
@nicholasjnelson let's chat after lunch about how to bring this up-to-date with the current master.

examples/custom_reference_states/MESA-input-1Msun-ZAMS.ipynb
```diff
@@ -0,0 +1,326 @@
...
+   "source": [
+    "file_write='cref_from_MESA.dat'\n",
+    "my_ref.write(file_write)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Now you can use this file to run a Rayleigh model. Once your Rayleigh model has run you can use the ``reference`` and  ``transport`` files to check how your specified reference state looks when transfered into Rayleigh."
```
gassmoeller: Can you provide an example input file in this folder that runs with the created files? That way users will have everything they need to run this in one place.

examples/custom_reference_states/mesa.py
```diff
@@ -0,0 +1,45 @@
+import numpy as np
+
+G = 6.67428e-8 # dyne cm^2 / g^2
```
gassmoeller: Question to @feathern: What is our policy regarding SI vs other unit systems? Do you want to keep everything uniform? Is everybody allowed to use their systems?

examples/custom_reference_states/reference_tools.py
```diff
@@ -0,0 +1,226 @@
+#Reference State Generation Tools  (reference_tools.py)
```
gassmoeller: Please move this file over to post_processing with `git mv examples/custom_reference_states/reference_tools.py post_processing/reference_tools.py`.

examples/custom_reference_states/mesa.py
feathern: I think we allow people to use whichever units.  At least half of us use cgs, which is completely sensible :)  Besides, the code is unit-agnostic.

gassmoeller: Not implying anything else :smile:. Ok then.

PR #157, opened 2019-07-24
Add link to CONTRIBUTING.md to documentation

gassmoeller:
This adds a link to the CONTRIBUTING.md document to the documentation. I could not figure out how to link to files outside of the doc repository without creating a link inside doc/source, but it seems this works well enough for now.


ljhwang:
The added file includes a softlink to the .md file in the repo. However, by doing it this way it requires an additional extension to be installed (probably why tester is failing).

Is there another way to do this?

gassmoeller:
I looked around some more, but I am pretty sure there is currently no better way. The only other options are:
- make a copy of CONTRIBUTING.md in the doc/source directory, which will easily get out of sync with the main file
- move the CONTRIBUTING.md file into the subdirectory, which makes it harder to find
- convert CONTRIBUTING.md to CONTRIBUTING.rst coud allow me to use a different directive, and would remove the requirement of recommonmark, but would require changing the file significantly

To me these alternatives look equivalent or worse than the current solution, but I am happy to change if you think one of them is better @ljhwang.

gassmoeller:
There is an open issue in sphinx to allow for relative links in toc-trees, but it is not resolved yet:
sphinx-doc/sphinx#701
There are some workarounds mentioned, but they do not work for our case.

gassmoeller:
This should be ready for merge.

doc/source/CONTRIBUTING.md
```diff
@@ -0,0 +1 @@
+../../CONTRIBUTING.md
```
cianwilson: Hadn't realized this was how github showed symlinks.  Took me a while to work that out!

gassmoeller: Yes I was confused as well when I uploaded them :smile:

PR #158, opened 2019-07-24
Add and use a reference section

### LINKED ISSUE #17: User Guide: Update Reference style
References throughout user guide should use a standard journal style. For example in the Overview section, reference as Glatzmaier (1984) and Clune et al. (1999).  If redundancy is desired within the text, then the full proper reference should be included with article name etc. This should be implemented so as to automatically generate a Reference section.

Any reference, asking to be cited should be included as a plain text bibtex (Section 9)

See sections: Overview, 3. Running a Benchmark, 4.3 Heimpel et al., 9. Acknowledging Rayleigh

ljhwang:
Adds and uses a reference section to the manual.  This addresses  #17 .
This keeps _authorname, YYYY_ style referencing inline for readability.
References are in a file separate from publications.

### DIFF
--- a/doc/source/User_Guide/benchmarking.rst
+++ b/doc/source/User_Guide/benchmarking.rst
@@ -8,19 +8,14 @@ Running a Benchmark
 ===================
 
 Rayleigh has been programmed with internal testing suite so that its
-results may be compared against benchmarks described in the following
-two papers:
-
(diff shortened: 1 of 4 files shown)


gassmoeller:
Looks nice, thanks!

PR #162, opened 2019-07-24
Added example for dimensional anelastic custom reference state

lkorre:
This is a jupyter notebook including an example of a custom reference state for a dimensional anelastic MHD setup in a convective spherical shell based on polytropes.


tukss:
Our automated testing finds that the notebook still contains the output cells. Can you run `make clear_ipynb` and amend your commit?

Issue #164, opened 2019-07-24
readthedocs Navigation menu out of sync with TO


ljhwang:
For the online documentation (https://rayleigh-documentation.readthedocs.io/en/latest/index.html ), navigation link _Outputs in Rayleigh_  looks like it is separate nav menu item, however, this is a line wrap problem.

PR #165, opened 2019-07-24
Elevate chapters in documentation

gassmoeller:
As discussed with @ljhwang this improves the order of sections and renames the label of one section to fit on the webpage.


ljhwang:
Thanks. This fixes issue #164 as well.

gassmoeller:
Thank you!

PR #173, opened 2019-07-26
Add a tester for convert_full3d_to_vtk.py.

cianwilson:
This piggybacks on the generic_input test by converting the full 3d output to vtk and then testing the contents of the resulting files match.


cianwilson:
I can't reproduce that failure even in a local build of that docker image.  Is it possible to trigger the build again in case it was a failure on the Jenkins server?

cianwilson:
The issue with the failing build turned out to be installing vtk in the Dockerfile through apt.  I now install it through pip and Jenkins seems happy.

cianwilson:
Thanks @gassmoeller !

PR #180, opened 2019-07-26
Add instructions for compiling the documentation

gassmoeller:
Add some instructions for how to compile the documentation.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -235,8 +235,26 @@ If the configure process fails to create Machine_Definitions, your build is a ch
 one.  Run './configure --blank' to generate a template that you may edit as needed.  
 IMPORTANT:  Ensure that whitespace is removed at the end of each line in this file.
 
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+V Building the documentation
 
+If you plan to modify the documentation you may want to build it locally to
+check your changes before submission. To do so you need a number of additional
+dependencies that are not required to build the Rayleigh executable. In
+particular you need the following packages:
 
+sphinx (including its dependencies)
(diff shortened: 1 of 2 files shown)


INSTALL
```diff
@@ -235,8 +235,26 @@ If the configure process fails to create Machine_Definitions, your build is a ch
...
+sphinx (including its dependencies)
+pandoc
+python
+numpy
+scipy
+sphinxcontrib-bibtex
+nbsphinx
+jupyter-client
+ipykernel
+
```
cianwilson: The `master` branch currently lists in `doc/requirements.txt` (in addition to those that are here already): recommonmark and matplotlib.  Should they be added to the list?

INSTALL
```diff
@@ -235,8 +235,26 @@ If the configure process fails to create Machine_Definitions, your build is a ch
...
-
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+V Building the documentation
+
+If you plan to modify the documentation you may want to build it locally to
+check your changes before submission. To do so you need a number of additional
+dependencies that are not required to build the Rayleigh executable. In
+particular you need the following packages:
+
+sphinx (including its dependencies)
```
cianwilson: What dependencies are being referred to here?  Don't all of the following have dependencies that aren't mentioned?

feathern: I agree.   For now, I would remove the parenthical statement.   I have some loose instructions on how to get a conda environment set up to visualize Rayleigh data.   I can update those to include Sphinx etc.   I think simply listing the base packages in the INSTALL file is sufficient for now.

gassmoeller:
Thanks for merging. You were right Cian, the instructions here were somewhat outdated (the documentation setup changed too quickly to keep them updated).

PR #181, opened 2019-07-26
Move doc source directory to .

gassmoeller:
Sphinx currently can not nicely link to files outside of the doc/ directory, which is why we had to create symlinks to directories in there. It looks like that is not fully working on readthedocs though (the plots of the python notebooks are not rendered correctly on the website). One way to solve this (and to make all files in Rayleigh available for the documentation) would be to use the main directory as the config directory for sphinx. This is just a proposal, I realize we might not want more files in the main directory, and I do not know if there could be conflicts down the line, but at the moment it seems to me like a nice solution.


gassmoeller:
Ok this is ready for a review. I actually like the new structure better than the old one.

gassmoeller:
Due to unknown reasons this does currently not build correctly on readthedocs (see https://gassmoeller-rayleigh.readthedocs.io/en/move_doc_source_directory/). The pdf builds fine, but the html does not generate an index page for readthedocs. Lets postpone this some time.

gassmoeller:
The current version works, but requires also moving `doc/source/index.rst` to `./index.rst`. Are we ok with that?

feathern:
@gassmoeller   I've approved this, but I can't seem to figure out what the merge conflicts are.

feathern:
Disregard my comment.  Somehow I had hosed my local copy of the Rayleigh repository.   Git status appears to be working after recloning it.  I can see what's actually conflicted now.

feathern:
OK, I was able to merge.   In addition to the git-recognized conflicts, there were some path-related conflicts with respect to [pr189](https://github.com/geodynamics/Rayleigh/pull/189), which came in and was merged while this request was awaiting review.    Specifically, examples.rst and diagnostics.rst referred to the old source/post_processing directory and needed to be adjusted.

PR #183, opened 2019-07-26
Allow some (T and C) bcs to be specified using generic spectral input

cianwilson:
This pull requests implements generic boundary conditions.  It makes a number of changes to achieve this.

First, boundary condition enforcement has been modified from a hand-rolled modification of the RHS arrays for each bc to a much more general boundary condition mask that gets initialized at start up and then applied every time-step to all RHS arrays.

Secondly, during that initialization step the mask can be set from generic spectral input files that are generated using `rayleigh_spectral_input.py`.  This involves adding new options `T_top_file`, `T_bottom_file`, `dTdr_top_file`, `dTdr_bottom_file`, `C_bottom_file` and `C_top_file`.  These override `T_top`, `T_bottom`, `dTdr_top`, `dTdr_bottom` and `impose_dipole_field`.  This can be expanded in the future but for now these options allow users to specify generic input files for temperature, radial temperature gradient and the poloidal magnetic field.  An example is provided in `tests/generic_input/bcs_script/main_input`.

Documentation has been added describing this functionality and the `tests/generic_input` test has been modified to test it.  This test runs rayleigh using hard coded boundary conditions and compares the shell slice output to the expected values and to a run using generic spectral input boundary conditions.  So this test achieves two goals, testing the new layout of the boundary condition enforcement and testing the generic input boundary conditions.

One tangential fix has been added to this pull request: initial condition file defaults are reset (just like the boundary condition ones) so that benchmarking mode works as expected.


orvedahl:
I won't claim to understand everything that was changed, but this looks good to me, including all the testing routines/results.

I did have two questions:
1) In the tests/generic_input/run_test.sh script, the first command changes directories as "cd tests/generic_input", which is where the script itself lives. Is this command necessary?

2) Is the src/Physics/Fields.F90_original file still needed or can that be removed?

cianwilson:
Thanks @orvedahl !

1. I guess that `cd` command isn't strictly necessary and could be moved lower down.  I wrote `run_test.sh` in a previous pull request when I thought it would get merged into the `Jenkinsfile` so I was mirroring the structure of the original test there that starts by `cd`ing too.  But then `run_test.sh` became too long and we decided it was better to leave it as a separate file.  That first `cd` could be merged with the `cd` a few lines down but it is necessary to cd  to run rayleigh in the right directory and I kind of like the symmetry or each sub-test cding from the main test directory where the output is also eventually tested by `test_output.py`.

2. As far as I can tell, no, though I'm guessing @feathern would have a better idea.  As it's unrelated to this pull request I've opened up another to remove it: #191 .

orvedahl:
That makes sense, I'm not too familiar with the Jenkins file stuff. I just commented out that first cd command and ran it from within the test/generic_input directory...I guess the cd should stay there if it was designed to be run from a different directory.

Either way, the more important part of this pull request looks good and I will approve it in just a second. I will leave merging it to someone else (mostly because github doesn't allow my work computer to merge; the button appears, but never does anything)

cianwilson:
Great, thanks @orvedahl !  I'm assuming I'm allowed to merge if it's been approved so I'll go ahead and press the button.

illorenzo7:
I believe this pull request is causing compilation failure on Pleiades, as well as the local JILA cluster (LCD for those familiar)--all use intel cores..  On the current master branch (and any point in git time >= this pull request #183) compilation will fail on Pleiades with the following message after trying to create the object file from Diagnostics_Base:

Diagnostics_Base.F90(174): error #6405: The same named entity from different modules and/or program units cannot be referenced.   [MY_R]
        Real*8, Intent(InOut) :: buffer(1:,my_r%min:,my_theta%min:,1:)
-------------------------------------------^
Diagnostics_Base.F90(174): internal error: Please visit 'http://www.intel.com/software/products/support' for assistance.
        Real*8, Intent(InOut) :: buffer(1:,my_r%min:,my_theta%min:,1:)
^
[ Aborting due to internal error. ]
compilation aborted for Diagnostics_Base.F90 (code 1)
Makefile:74: recipe for target 'Diagnostics_Base.o' failed
make[3]: *** [Diagnostics_Base.o] Error 1
Makefile:58: recipe for target 'dbg' failed
make[2]: *** [dbg] Error 2

I am not sure why Diagnostics_Base is causing the issue, since that hasn't been modified in ~1 year. I also can't find a place where my_r is referenced in the lines of code changed by #183. My understanding of "git time" could also be wrong, in which case another pull request is creating the problem. When I check out

0952a48 (#157) compilation succeeds

but on 
93d4802 (#183) compilation fails with the above message

#193 and #198 also give me this error. 

It sounds like this is particular to the Pleiades/LCD environment, however, since clearly the pull request compiled on the Jenkins server and I don't think other people have been facing compilation issues. I am not sure what in those environments changed--any insight anyone has would be much appreciated, since I won't be able to work off the master branch until this is resolved!

orvedahl:
I noticed the same compile issue earlier today on CU's supercomputer, and I used the Intel compiler. The exact same version of Rayleigh compiled on my work station with the GCC compiler, so I think this is a compiler issue, not a compute-environment issue.

I did notice the variable "my_r" is defined in the ProblemSize.F90 module as well as the Load_Balance.F90 module, but I'm not sure what implications that has.

feathern:
I don't think this is a compiler issue.  I think that Intel and GNU tolerate circular module references differently.  This error looks like a circular module reference to me.  I'm looking at this now.  Hopefully I can sort this out tomorrow.  Someone please be ready to review so we don't have to wait a week ;)

feathern:
I have a fix -- will post tomorrow.   This  one's my fault.  Cian's the first person besides me to need to use the Load_Balance module, and there's some rough edges in there that need to be smoothed out.

PR #184, opened 2019-07-26
added python notebooks for volume rendering with YT

mariaweber:
Added two new notebooks for playing around with volume rendering in YT. Comments in the notebook indicate where there are still issues with YT capabilities that need to be sorted out.

-Tutorial1 provides an overview
-Tutorial2 shows some additional examples and shows how to convert from Spherical quantities to Cartesian quantities within Python.


mariaweber:
@nicholasjnelson I believe @feathern will be including a low-res sample from a checkpoint to use in the notebooks in the future.

PR #186, opened 2019-07-26
added new directory with code for cartesian interpolation

mariaweber:
Included a new directory with Kyle's parallel interpolation code (with input and Makefiles) to interpolate Spherical_3D quantities onto a Cartesian grid. I can confirm that these compile and work as expected to generate the 'cubed' output files for use with visualizations.

### DIFF
--- a/post_processing/kyle_interp/Makefile
+++ b/post_processing/kyle_interp/Makefile
@@ -1,0 +1,53 @@
+# CU/NCAR Janus
+DOPTS = 1
+
+.SUFFIXES: .o .f90
+
+MPIF90 = ifort
+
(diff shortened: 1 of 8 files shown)


tukss:
If @kyle-augustson is OK with this, we can add the GPLv3 notices at the top of the source files.

cianwilson:
This looks good, though I haven't looked at it in detail.

Perhaps it could be given a more descriptive folder name than `kyle_interp`?  Since Kyle isn't the one committing it perhaps his authorship could be acknowledged with a `README` file in that directory rather than in the directory name.

feathern:
I agree.   I think I may have actually suggested "kyle_interp," but let's just call it something more descriptive, like '3d_interpolation' and drop a README in there.   I am going to go through and clean this + the Makefiles up once Maria has made the initial commit and can do that then.

p.s.  Just to confirm, we do have Kyle's permission to drop this into the repository.

PR #189, opened 2019-07-27
Fixes path to plotting examples.

ljhwang:
This fixes references to the plotting notebook from Diagnostic Outputs. The notebook can now be accessed from navigation after References with the link "Plotting Examples" as well as access through the section Diagnostic Outputs. Note the we do not keep pdf's anymore so references elsewhere need to be updated still.


ljhwang:
Thanks!

There are known WARNING messages when makie’ing the file related to *i.pynb.
We’re still thinking about how to fix this.


Best,
-Lorraine

*****************************
Lorraine Hwang, Ph.D.
Computational Infrastructure for Geodynamics
530.752.3656

PR #192, opened 2019-07-29
Allow generic input to be read back into python and improve reading and writing

cianwilson:
The `SpectralInput` class of `rayleigh_spectral_input.py` can now read files back in using the `read` function.  This is mostly useful for debugging as the files are binary and unreadable except by rayleigh otherwise.

I've also made `SpectralInput` honor endianness through byteswapping much like is done in `pre_processing`.

I've added a test for this functionality in `tests/generic_input/test_output.py`.


cianwilson:
That's just in the test.  I needed something to compare the read-in file against so I just transformed a constant function that returns 5.  A bit over the top but at the moment there's only an interface for transform_from_rtp_function.  Could have also done it by just knowing the answer for such a simple case :-)

The other 5.0 is because I decided to test the inverse_transform function.  It's not efficiently implemented but it's useful for debugging.  So I inverse transform the read-in file and check that it still thinks it's 5.

cianwilson:
Thanks @feathern!

PR #195, opened 2019-07-30
Custom reference interface

orvedahl:
This pull request allows you to use one of the existing reference types and slightly modify it using the custom reference files. The diffusivities can be changed, as well as the buoyancy and heating terms.

Nick is working on the documentation for this capability as well as some notebook examples. Those will be added in a separate pull request.


feathern:
I just want to reiterate what Ryan said.   I am hoping to get the documentation for this drafted today.    If we'd like to hold off on the actual merge until the docs are ready, that's OK by me, but getting a review going on this sooner rather than later could be very helpful.
-Nick

feathern:
I took care of the conflict in reference_tools.py.   The change in the equation_coefficients class will break some of the custom reference notebooks.  I can fix those in a separate pull request once this is approved and I can easily branch off an up-to-date copy of the code.

feathern:
Loren and I had a long email conversation.  I thought he was either going to come by my office this morning, but he hasn't shown up.  Here's his email below.  I'm going to go ahead and merge this.
> Currently if I take f_6 / (f_1*f_4) from the output equation_coefficients file I get the same thing that used to be ref%heating in the old code, namely a constant ~6 x 10^(-5) (for solar CZ case), which is right for constant-entropy heating (heating_type=1). So f_6 = Q(r), which integrates over the whole shell to the solar luminosity. This is also consistent with c_10 = 1.

> I do think it's strange that constants mostly get set to 1...of course you could always choose to define them as one with the functions absorbing them, but I feel like the output should reflect the equation system as it was input. Also, the equation_coefficients data file always has cset and fset hard-coded to 1 (this happens on lines 848-849 in Write_Equation_Coefficients() in ReferenceState.F90). This might make sense if they are supposed to specify whether the functions and constants are ultimately set--whether in a custom_reference file or by the built-in routines. And they should better be set ultimately somehow. But then it seems like the equation_coefficients structure is just carrying around some extra arrays of ones. 

> Other than how it's outputting the file equation_coefficients, I do think the behavior is as intended, so I'd be good to merge it as is and talk about how to output equation_coefficients later. But I can wait if you'd prefer.

PR #197, opened 2019-07-30
Multi-domain L conservation weights

connorbice:
Updated the L conservation weights used with the strict L conservation boundary condition to be valid for runs with multiple Chebyshev domains.

### DIFF
--- a/src/Physics/BoundaryConditions.F90
+++ b/src/Physics/BoundaryConditions.F90
@@ -61,7 +61,6 @@ Module BoundaryConditions
     Character*120 :: C_top_file       = '__nothing__'
     Character*120 :: C_bottom_file    = '__nothing__'
 
-    Logical :: Gentle_L_Conservation = .false.
     Logical :: Strict_L_Conservation = .false.         ! (In-Progress) Turn on to enforce angular momentum conservation abous x,y, and z-axes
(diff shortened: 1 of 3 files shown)


connorbice:
[Here](https://drive.google.com/open?id=1Zt7aexbdeCs47mQpuMlnT-YHvPKj4x6m) is the verification figure for the same model split into 1, 2, and 3 Chebyshev domains.

connorbice:
And a [verification figure](https://drive.google.com/open?id=1YXa3xVNQ6fifREguWNCY8F3cyx9HbuVO) for my new gentle_L_conservation. Apologies for slipping it in with the other change, I didn't realize it would turn into the same pull request when I commited it.

feathern:
Hey Connor,
This looks good, but I'd like for you to remove the gentle_L_conservation stuff for now.   It looks as though you are subtracting the accumulated error from the n=0 Chebyshev mode, but  the radial integration involves a volume integral.  I'm pretty sure that the factor of r^2 associated with the volume element means that other modes besides n=0 survive the averaging, and so the net angular momentum can project onto other modes as well.  I may just be confused, and I'm happy to be convinced otherwise by some rigorous math, but for now, can we keep this PR to the strict_L_conservation flag only?     I agree that it appears to be implemented correctly.
-Nick

connorbice:
GLC stuff has been removed.

PR #199, opened 2019-08-01
Loadfix

feathern:
This fixes the conflicting references to my_r, defined in problemsize.F90 and in Load_Balance.F90.  While they should have the same value, they are not the same variable.   For now, when using Load_Balance.F90, the 'Use Only' syntax needs to be used.   I plan to provide an update later where my_n_lm etc. are properly stored in the parallel interface class and hten passed into problemsize.F90.  For now, this will fix conflicts.   Also, conf.py had a small whitespace conflict on my branch.   Should be safe to use this version.
@orvedahl @illorenzo7 @cianwilson   Could one of you put eyes on this quickly, since the bug is preventing compilation on Pleiades?  Thanks!


feathern:
Quick note.  Fields.F90 doesn't need to see Problemsize.F90, but removing that dependency without running makedepf90 (which I haven't learned to use yet) caused a problem, so I put it back for now.  The Use statement is innocuous enough.   Thanks for the approval.  I will merge.

PR #203, opened 2019-08-02
Removed all references to the STABLE plugin

feathern:
This PR removes all references to the STABLE plugin.  That functionality remains available in the rayleigh-0.9.5 branch created at the beginning of the workshop.   STABLE was written by M. Miesch and B. Karak and used in a single paper.  There is  no documentation on how to use it, and it would be better implemented as a patch later, I think.
-Nick


feathern:
Not sure why the automated test isn't running, but this compiled fine on my machine...

PR #204, opened 2019-08-02
Merge ReferenceState.F90 and TransportCoefficients.F90

feathern:
This PR merges ReferenceState.F90 and TransportCoefficients.F90 into a single module named PDE_Coefficients.F90.  Routine names and variable names remain largely unchanged.  Main_input files do not require modification to use this new version.

This is a large module, but it provides a single space within which all PDE coefficients can be manipulated.  Several bits of housecleaning were performed in the conversion.   I have run the Christensen 2001 hydo benchmark and things seem to be working.


feathern:
Again, not sure what's going on with the automated tester.  Maybe the server is down?

feathern:
@gassmoeller I see a conflict with Makefile.deps now.   I'm wondering if somehow the issue leading to your earlier PR came through me somehow?

feathern:
@cianwilson  -- Any chance you could look at this sometime in the next week?   This is probably a good one for you.
-Nick

gassmoeller:
The conflict is there because this PR fixes the same line in Makefile.fdeps as #205. When you resolve the conflict just make sure the Run_Param_Header.opt.o line stays in there.

feathern:
OK -- just fixed the merge conflict.

orvedahl:
I have three comments/questions:

1) The Write_Reference subroutine is still there and still used, but the one that writes the transport file is no longer there. Should they both be removed in favor of the equation_coefficients file?

2) The reference_tools.py file has two versions of the gen_poly() function, the compute_heating_profile() function, and the background_state class.

3) The rayleigh_diagnostics.py file still has the ReferenceState and TransportCoeffs. The current ReferenceState will not read the currently written reference file and the transport file is not even written anymore. Should these classes be merged so that they read the equation_coefficients file?

feathern:
Hey Ryan,
Point 1:   Good point - I should remove write_reference
Points 2 and 3:   I agree that those python files need to be modified, but nothing in this commit really affects those (they were a little out of whack already).  I would suggest documenting those problems into an issue that someone can have a look at.  The purpose of this commit was really just to create a unified module to help, in party, to lay the framework for  the new checkpointing.

feathern:
OK, write_reference has been removed.

PR #206, opened 2019-08-06
Added -mac-mkl flag to configure

feathern:
This commit adds support for linking the MKL library on Macs.  Presently, only conda-installed MKL is supported.     This is configuration is enabled by passing the -mac-mkl flag to configure.  In addition, I changed things around so that the -conda-mkl flag can be used without the -mkl flag.  Initially, they had to be used to together, but that felt redundant.

The reason for these conda flags is that I am planning to add some documentation on setting up a conda environment for building Rayleigh and the docs.

-Nick

### DIFF
--- a/configure
+++ b/configure
@@ -158,11 +158,17 @@ function displayhelp {
     echo ""
 
     echo "  -conda-mkl:     " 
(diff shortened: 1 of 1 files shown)


feathern:
Going to go ahead and merge this.   Don't seem to be many takers...

PR #208, opened 2019-08-06
First draft of conda instructions.

feathern:
This PR adds documentation describing how to set up a single Conda development environment that allows you to compile both Rayleigh and its documentation.

In order for these instructions to work completely, PR 206 needs to be approved.

In the meantime, if someone could test this, that would be great.  Just add an extra -mkl flag to the flags already specified for configure.  I will check for typos tomorrow or correct them soon if this is merged quickly.

-Nick

### DIFF
--- a/doc/source/User_Guide/devel_environment.rst
+++ b/doc/source/User_Guide/devel_environment.rst
@@ -1,0 +1,123 @@
+Setting up a Rayleigh Development Environment
+==============================================
+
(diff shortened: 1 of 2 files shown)


feathern:
Quick note.   The -conda-mac flag won't work until that other pull request is approved.   -mkl -conda-mkl will work on Linux, though.

PR #209, opened 2019-08-08
Add class to read equation_coefficients file

orvedahl:
This pull request removes the ReferenceState and TransportCoeffs classes from the rayleigh_diagnostics.py since the reference file and transport files are no longer written. It also removes various doubly-defined functions and classes in the reference_tools.py file.

The removed classes have been replaced with PDE_Coefficients. This class reads the new equation_coefficients file and extracts the constants and functions. It also has some aliases to frequently used functions, such as density, temperature, nu, kappa.

I have adjusted the example notebooks to point to the new class.


feathern:
Hey Ryan,
I was talking with Loren about this a little, and we think that the ReferenceState and TransportCoefficient classes should stay in rayleigh_diagnostics.py so that people can read their old outputs.  Can you add them back in?  I will try to have a more detailed look at this tomorrow.
-Nick

orvedahl:
I put back the ReferenceState and the TransportCoeffs. The ReferenceState should read all three versions of the reference file; the previous ones with or without heating and the most recent one that no longer has entropy/gravity/pressure.

PR #210, opened 2019-08-08
Fixed reference-state advection bug.

feathern:
This PR fixes a bug where reference-state advection was improperly implemented.   The term v dot grad dsdr is linear and implemented semi-implicitly through the CN scheme.   As it stood, only the implicit portion (left-hand side) was implemented.  The explicit portion (right-hand side) had been omitted.   The RHS is now properly addressed through this PR.


feathern:
Looks like the tester is down.   I'll give this until tomorrow before merging, but I compiled the code and the conditional is pretty straighforward, so I think it's OK.

PR #211, opened 2019-08-09
Made sure diffusion functions and constants are properly output to eq…

illorenzo7:
…uation_coefficients file

### DIFF
--- a/src/Physics/PDE_Coefficients.F90
+++ b/src/Physics/PDE_Coefficients.F90
@@ -1148,9 +1148,6 @@ Contains
 		    ra_functions(:,8) = ref%dlnrho(:)
 		    ra_functions(:,9) = ref%d2lnrho(:)
 		    ra_functions(:,10) = ref%dlnT(:)
-		    ra_functions(:,11) = dlnu(:)
-		    ra_functions(:,12) = dlnkappa(:)
-		    ra_functions(:,13) = dlneta(:)
 		    ra_functions(:,14) = ref%dsdr(:)
 		Endif
         filename = Trim(my_path)//"equation_coefficients"
--- a/src/Physics/PDE_Coefficients.F90
+++ b/src/Physics/PDE_Coefficients.F90
@@ -1014,11 +1014,11 @@ Contains
     End Subroutine Allocate_Transport_Coefficients
 
     Subroutine Initialize_Diffusivity(x,dlnx,xtop,xtype,xpower,ci,fi,dlnfi)
-	    Implicit None
-	    Real*8, Intent(InOut) :: x(:), dlnx(:)
-	    Real*8, Intent(InOut) :: xtop
-	    Integer, Intent(In) :: ci, fi, dlnfi, xtype
-	    Real*8, Intent(In) :: xpower
+        Implicit None
+        Real*8, Intent(InOut) :: x(:), dlnx(:)
+        Real*8, Intent(InOut) :: xtop
+        Integer, Intent(In) :: ci, fi, dlnfi, xtype
+        Real*8, Intent(In) :: xpower
         Character(len=2) :: ind
 
         If (reference_type .eq. 4) Then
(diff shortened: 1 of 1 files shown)


orvedahl:
I'm not convinced this is necessary, compared to what was already in the code.

And more importantly, you are forcing constant diffusivities unless a dimensional polytrope is used. Boussinesq is allowed to have non-constant diffusivities and there's no reason the non-dimensional polytrope should be any different than the dimensional one. The process of non-dimensionalizing is simply a rescaling of the function values.

feathern:
Hey Loren,

Thanks for this.  This is close to what we want to do, I think, but I have a few comments.  

1.  The following bit of code is unnecessary because ra_functions and ra_constants are initialized to zero.  Plus, a  user might want to supply a custom reference state initially for a hydro run and later for an MHD run.  The code below would prevent and eta profile and eta top from being saved in subsequent checkpoints.   Since magnetic features are enabled through the logical magnetism flag, and we don't need to set these values to zero if magnetism is off, I would suggest we remove this. 

`
ra_functions(:, 7) = 0.0d0
ra_constants(7) = 0.0d0
`

2.  Next, I want to mention that nondimensional modes ARE intended to allow variable diffusivities.  In particular, lots of users want to apply variable eta.  Eta_top, nu_top, and kappa_top are set in the nondimensional routines now, but the actual profile can be varied.  For this reason, I don't think the logic at the end of intialize_diffusivity is what we want to do.


3.  I think that rather than basing how we set ra_constants and ra_functions on the reference state type, we need to do it on [nu,kappa,eta]_type.   What do you think about the code below?  I think this does what we want in every instance with the possible exception of the elseif in case(3).   In that case xtop is not longer really the top value.   It's the prefactor that multiplies the user's function.  I think this is what we probably want to do, but we need to document the behavior of xtop in this instance.
`

        Select Case(xtype) 
            Case(1)  
                x(:) = xtop
                dlnx(:) = 0.0d0
                ra_constants(ci) = xtop
                ra_functions(:,fi) = 1.0d0
                ra_functions(:,dlnfi) = 0.0d0
            Case(2)
                Call vary_with_density(x,dlnx,xtop, xpower)
                ra_constants(ci) = xtop
                ra_functions(:,fi) = x(:)/xtop
                ra_functions(:,dlnfi) = dlnx
            Case(3)
                If ((ra_function_set(fi) .eq. 1) .and. (ra_constant_set(ci) .eq. 1)) Then
                    x(:) = ra_constants(ci)*ra_functions(:,fi)
                    dlnx(:) = ra_functions(:,dlnfi)
                    xtop = x(1)
                    ! Nothing to be done here for functions and constants -- completely set
                ElseIf ((ra_function_set(fi) .eq. 1) .and. (ra_constant_set(ci) .eq. 0)) Then
                    x(:) = xtop*ra_functions(:,fi)
                    dlnx(:) = ra_functions(:,dlnfi)
                    ra_constants(ci) = xtop
                Else
                    If (my_rank .eq. 0) Then
                        write(ind, '(I2)') fi
                        Call stdout%print('Error: Need to specify f_'//Trim(ind))
                    EndIf
                EndIf
`

illorenzo7:
The reason something is necessary is that currently (for example) c_6 is always set to 1. According to the documentation, this should only be true for the dimensional polytrope. For the Boussinesq case, c_6 = 1/Pr, and for the ND anelastic case, c_6 = E/Pr. In both non-dimensional cases f_5 = 1. I agree that once you take the product c_6*f_5 it doesn't matter (mathematically) whether the function is 1 or the constant is 1. But I do think that what's output in the equation_coefficients file should reflect what is input in the code, which presumably is what is in the documentation.

In the equation sets solved by Rayleigh (also according to the documentation) all diffusivities are assumed to be constant. I don't know what happens if you chose ref_type = 1 (for example) and then choose a radially varying diffusivity but according to the documentation this case doesn't seem to be supported. But it's very possible I'm confused. Perhaps @feathern can comment.

feathern:
I agree with @illorenzo7 -- there are several errors in the way we're outputting the equation coefficients.  At least two constants are having their values lost.  Also, as Loren pointed out, it would be best if we could have the equation coefficients output reflect as closely as possible what the user puts in.  Please see my comments above!
@orvedahl  Loren and I had a long talk about this yesterday.  There are several other small commits he's going to make soon to try to get this working as well as possible for all 4 reference types.
Thanks Guys!
-Nick

illorenzo7:
@feathern I agree with point 1, I hadn't realized that ra_constants and ra_functions were initialized to zero. 

Regarding 2 and 3, I think this might be a documentation issue, which had me confused. In the documentation, both ND cases have f_3, f_5, f_7 --> 1(everywhere) and c_5, c_6, c_7 --> xtop. Also in the equations listed, the diffusivities are assumed to be constant in both ND cases. It sounds like this is not the case? If the diffusions ARE allowed to vary in the Boussinesq case, for example, the equations should be updated, and it sounds like the mapping of constants/functions listed in the documentation is also incorrect. 

Regarding your code, it looks fine as long as it actually maps to the equations solved by Rayleigh! It doesn't seem to map onto what's listed in the documentation however, since there whether or not c_5, c_6, c_7 = 1, for example, depends on the reference type chosen, not on the transport type, as in the code above.

feathern:
I agree the documentation needs to be updated and/or clarified.  I can look into that once we have all of this finalized.   It's mostly an issue of the documentation being out of date.  There was some discussion at the workshop, and we decided that people should be able to freely specify a diffusivity profile for all reference types, so this is a relatively new feature.  It's discuss a little in my custom reference documentation, but you wouldn't see that when looking at "the equations solved by Rayleigh."   I will start giving some thought to overhauling the organization of those parts of the docs.

illorenzo7:
Okay, pending documentation updates, I trust Nick! I've implemented his fix to how the functions/constants are set and removed the unnecessary lines having to do with the eta constant/function.

PR #214, opened 2019-08-09
direct attribute access in equation_coefficients

tukss:
This allows us to directly reference functions and coefficients by name
from the equation_coefficients class.
    
Example:
```python
cref = rt.equation_coefficients(file='equation_coefficients')
cref.density
```

### DIFF
--- a/post_processing/reference_tools.py
+++ b/post_processing/reference_tools.py
@@ -6,6 +6,12 @@ class equation_coefficients:
     nconst = 10
     nfunc = 14
     version=1
+    c_dict    = {'two_omega':1, 'buoy_fact':2, 'p_fact':3, 'lorentz_fact':4, 'visc_fact':5,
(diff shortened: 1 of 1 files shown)


feathern:
Hey Philipp,
Can you comment on why you changed the last line from y[:] to y.   Does this not create a pointer to y, or is the slice on the LHS sufficient to perform a copy?
-Nick

tukss:
> Can you comment on why you changed the last line from y[:] to y. Does this not create a pointer to y, or is the slice on the LHS sufficient to perform a copy?

The slice on the LHS is sufficient to cause a copy here. The advantage of not using a slice in the RHS is that we can pass scalars to it.

Also a slice on the RHS actually doesn't create a copy but a mutable "view" on the other array. Consider this code:
```python
import numpy as np
A=np.zeros((2,2))
b = A[:, 0] # b is not a mutable view on a column slice of the A array.
b[:] = 1 # This overwrites the original location b is "viewing".
print(A)
# [[1. 0.]
# [1. 0.]]
```

PR #215, opened 2019-08-09
Ref mod

feathern:
This is a small modification that will prevent augment_reference_state from overwriting ra_functions and ra_constants as set by ref_type=1,2,3.  For the moment, this isn't an issue, but @illorenzo7 is working on separate commits where the functions and constants are set within each of those routines.   

A similar problem exists in relation to the transport coefficients.   I will fix that in a similar manner once this PR has been examined and approved.


feathern:
Thanks @illorenzo7.   I have the other addition ready, but I don't appear to have pushed the branch to my git repo.  I'll push it once I get home.

PR #216, opened 2019-08-12
Set equation coefs reftype1

illorenzo7:
This is the second in a series of commits to properly set the equation coefficients in the "equation_coefficients" binary file. The first dealt with the diffusion coefficients, (c_5, f_3), (c_6, f_5), and (c_7, f_7) and was approved. This deals with the other c's/f's in the non-dimensional boussinesq case (reference_type = 1 and the Constant_Reference() subroutine).

### DIFF
--- a/src/Physics/PDE_Coefficients.F90
+++ b/src/Physics/PDE_Coefficients.F90
@@ -405,7 +405,7 @@ Contains
     Subroutine Polytropic_Reference()
         Real*8 :: zeta_0,  c0, c1, d
         Real*8 :: rho_c, P_c, T_c,denom
-    Real*8 :: thermo_gamma, volume_specific_heat
+        Real*8 :: thermo_gamma, volume_specific_heat
         Real*8 :: beta
         Real*8 :: Gravitational_Constant = 6.67d-8 ! cgs units
         Real*8, Allocatable :: zeta(:), gravity(:)
@@ -503,7 +503,7 @@ Contains
         Else
             ref%Lorentz_Coeff = 0.0d0
             ref%ohmic_amp(1:N_R) = 0.0d0
-        Endif   
+        Endif 
 
     End Subroutine Polytropic_Reference
 
--- a/src/Physics/PDE_Coefficients.F90
+++ b/src/Physics/PDE_Coefficients.F90
@@ -400,27 +400,6 @@ Contains
             ref%ohmic_amp(1:N_R) = 0.0d0
         Endif
 
-        ! Set the equation coefficients (apart from the ones having to do with diffusivities and heating)
-        ! for proper output to the equation_coefficients file
-        ra_functions(:,1) = ref%density
-        ra_functions(:,2) = gravity*ref%density
-        ra_functions(:,4) = ref%temperature
-        ra_functions(:,8) = ref%dlnrho
-        ra_functions(:,9) = ref%d2lnrho        
-        ra_functions(:,10) = ref%dlnT
-        ra_functions(:,14) = ref%dsdr     
-                        
-        ra_constants(1) = ref%Coriolis_Coeff
-        ra_constants(2) = Modified_Rayleigh_Number
-        ra_constants(3) = 1.0d0
-        ra_constants(4) = ref%Lorentz_Coeff
-        ra_constants(8) = Ekman_Number*Dissipation_Number/Modified_Rayleigh_Number
-        If (magnetism) Then
-        	ra_constants(9) = Ekman_Number**2*Dissipation_Number/(Magnetic_Prandtl_Number**2*Modified_Rayleigh_Number)
-        Else
-        	ra_constants(9) = 0.0d0
-        Endif
-
     End Subroutine Polytropic_ReferenceND
 
     Subroutine Polytropic_Reference()
@@ -524,24 +503,7 @@ Contains
         Else
             ref%Lorentz_Coeff = 0.0d0
             ref%ohmic_amp(1:N_R) = 0.0d0
-        Endif
-        
-        ! Set the equation coefficients (apart from the ones having to do with diffusivities and heating)
-        ! for proper output to the equation_coefficients file
-        ra_functions(:,1) = ref%density
-        ra_functions(:,2) = ref%Buoyancy_Coeff
-        ra_functions(:,4) = ref%temperature
-        ra_functions(:,8) = ref%dlnrho
-        ra_functions(:,9) = ref%d2lnrho        
-        ra_functions(:,10) = ref%dlnT
-        ra_functions(:,14) = ref%dsdr     
-                        
-        ra_constants(1) = ref%Coriolis_Coeff
-        ra_constants(2) = 1.0d0
-        ra_constants(3) = 1.0d0
-        ra_constants(4) = ref%Lorentz_Coeff
-        ra_constants(8) = 1.0d0
-        ra_constants(9) = ref%Lorentz_Coeff       
+        Endif   
 
     End Subroutine Polytropic_Reference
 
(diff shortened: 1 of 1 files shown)


src/Physics/PDE_Coefficients.F90
```diff
@@ -454,9 +473,9 @@ Subroutine Polytropic_Reference()
         ! Initialize reference structure
         Gravity = Gravitational_Constant * poly_mass / Radius**2
 
-	! The following is needed to calculate the entropy gradient
-	thermo_gamma = 5.0d0/3.0d0
-	volume_specific_heat = pressure_specific_heat / thermo_gamma
+		! The following is needed to calculate the entropy gradient
```
tukss: This seems to be an unnecessary change in whitespace. Also tabs in Fortran are sometimes problematic. ;-) Maybe just remove that part from the commit to clean it up.

src/Physics/PDE_Coefficients.F90
```diff
@@ -484,7 +503,7 @@ Subroutine Polytropic_Reference()
         Else
             ref%Lorentz_Coeff = 0.0d0
             ref%ohmic_amp(1:N_R) = 0.0d0
-        Endif
+        Endif
```
tukss: Same goes for the whitespace change here.

feathern:
I agree - no tabs :)   I will look at this soonish - probably on Wednesday - and after Loren has fixed the indexing issue.

illorenzo7:
Thanks @tukss. Just out of curiosity, where were you seeing at out-of bounds error? The line in question was definitely wrong logically, so I've fixed it. But what seemed to be happening is that i = N_R from the prior loop so it would have just been setting ra_functions(:,2) all equal to 1--I'm not sure how "i" (not I!) would have fallen outside the range 1, N_R. 

In any event, it should be fixed now. 

I also have fixed the whitespace issue (I think). Vim expands tabs to 4 spaces, like a reasonable text editor might be expected to do. Gedit seems to "expand" tabs to tabs of width 4 spaces, like a *****. These tabs (since they are really tabs) don't stay 4 spaces anywhere but my machine, apparently, to trick me into thinking they are not there. I'll be more careful in the future, and use vim :-)

tukss:
> Thanks @tukss. Just out of curiosity, where were you seeing at out-of bounds error?

You see the out of-bounds error when you click the link next to the message about the failed test. It takes you to the Jenkins page that shows output of the tests, which we fortunately run with the array boundary checker on.

> The line in question was definitely wrong logically, so I've fixed it. But what seemed to be happening is that i = N_R from the prior loop so it would have just been setting ra_functions(:,2) all equal to 1--I'm not sure how "i" (not I!) would have fallen outside the range 1, N_R.

When you run a `do` loop in Fortran, the iteration variable will be one beyond the last value. This is even guaranteed by the standard. It is a good way to check if the loop went through completely or it was prematurely finished with an `exit` statement. The classical `for (i=0; i<n; i++)` loop in C is quite similar in that respect.

tukss:
Looks good to me now. Maybe @feathern should have a quick look at the coefficients before we merge.

Somehow I don't see the Jenkins tests being run now. Maybe there is some issue with the Jenkins server again?

illorenzo7:
I see now how the out-of-bounds error came about. Thanks @tukss, I had forgotten that about C/Fortran loops!

orvedahl:
I may be reading this wrong, but should function_6 and constant_10 be set based off the ref%heating that was set in the initialize_reference_heating? As it is currently written, if the user has heating_type=2, then the ref%heating will be properly set, but it will still write f_6 and c_10 as zero.

illorenzo7:
@orvedahl I was wondering the same thing. Currently I am going off of what's in the documentation, which has reference_type = 3 implying heating_type = 0, i.e., f_6 = 0, c_10= 0, and no Q appears in the equation for temperature evolution. 

Nick and I are also working on the documentation of the equation set separately. I think the plan is to move forward setting the equation_coefficients based on the documentation as it stands, even though we know some parts are wrong. Then we can fix those parts after fixing the documentation. 

Currently, I believe the only thing in the documentation that's glaringly wrong and/or unspecified are the conventions for the f's and c's relating to heating and the transport coefficients, as well as the nondimensional anelastic case when there is no rotation. I don't think these conventions have been decided upon, though we have the current version for the transport coefficients using the code Nick suggested. 

So, for now, I think we all just have to accept that the equation_coefficients file is not going to contain correct information regarding the heating and transport coefficients. Hopefully most people only need the temperature, density, etc., which should still be correct. Otherwise, just know what you put in!

@feathern Ryan's comment is making me think again we should wait on my series of pull requests until we sort out what we want all the conventions on the f's and c's to be first (i.e., fix the documentation). But what do you think?

illorenzo7:
...sorry I meant reference_type = 1, not 3.

illorenzo7:
Sorry for the multiple comments: regarding the heating, I do have a pull request ready to go to to set c_10 and f_6 in the initialize_reference_heating() routine, which can easily include the Boussinesq case. Then it would make most sense to remove the setting of c_10 and f_6 in this pull request. However, I am using "my" convention for Q(r) = c_10*f_6, where c_10 equals the luminosity or heating_integral, and f_6 is a radial "shape" that integrates over the full shell to 1. So before I submit the heating pull request, I need to if 1) nonzero heating type is supported when reference_type=1 and 2) whether "my convention" on Q(r) coheres with what everyone wants the actual convention to be.

orvedahl:
In the current documentation under the reference_type=1 section, it gives a list of the functions and constants under the given nondimensionalization. Then the paragraph just after that says in the middle "We might choose to specify fixed-flux boundary conditions and/or internal heating through a suitable choice f_6(r), in which case Ra refers to a flux-based Rayleigh number"

So I think the current documentation is correct, but the equations that are displayed are for a specific choice, which is to say, heating_type=0 and constant diffusivity.

illorenzo7:
Right, and I was confused by that paragraph :-) The problem is, I don't know what goes into a non-dimensionalization where heating_type doesn't equal 0; does it only change f_6 and c_10 to something nonzero, or will it change other parts of the energy equation since effectively it is enforcing a different temperature scale? I also wasn't sure if this paragraph was ``left over" from something else and so didn't apply. Normally, Boussinesq models exclude any forms of heating. Also, how is the heating non-dimensionalized? This choice will alter other f's and c's. 

For the current pull request, I see two options:
1) remove any setting of f_6 and c_10, and we deal with it later
2) add a flag setting the ra_constants and fa_functions ONLY if heating_type = 0, effectively minimizing the impact of this pull request. 

I would personally favor option 2, since it sounds like the current mapping of f's and c's in the documentation only applies when heating_type = 0. 

@feathern I am starting to think these pull requests will all be so much easier once I know what the c's and f's actually are for each case, rather than dealing with it later. ...Or someone else should do the pull requests who knows what the f's and c's are!

feathern:
Sorry, I was pretty busy yesterday.

First, a general note.  The documentation was written fairly rapidly, and was meant to give a quick overview of the three reference types in Rayleigh.  It wasn't necessarily meant to be exhaustive, and it clearly needs some updating/clarification.   With that in mind, here are my two general comments.

1.  f6 and c10 should be set within the initialize reference heating routine.   Someone running the standard Boussinesq setup may want to have a heating function.   See for instance, this paper:  https://arxiv.org/abs/1210.8154   This is why initialize_reference_heating is called in constant_reference.    f6 and c10 shouldn't be assumed to be zero.  They often are if ref_type=1 because most people fix temperature on the boundaries and have no internal heating (hence the documentation's wording).

2.  The earlier suggestion of f6 being radial shape that integrates to 1 and c10 being luminosity or heating integral is sensible.  I suggest doing that -- in fact, I think it's why I added the heating_integral variable in the first place.

2. Regarding the Rayleigh number question/comment.   Turning on internal heating will typically (maybe always, but there are always exceptions) change the meaning of the Rayleigh number from the standard definition to a flux Rayleigh number.   BUT, so would changing the boundary conditions to fixed flux on both boundaries, or fixed flux on one boundary and fixed T on the other.   The equations appear the same, but the temperature scale chosen is different.   If F is the imposed flux, F ~ kappa*rho*cp*T/L  (where L is shell depth).   The new temperature scale, for Boussinesq setup would be T = F*L/(kappa*rho*cp).   The only thing that actually changes when you nondimensionalize is the prefactor in front of the heating function.  Everything else looks the same, but the understanding of the Rayleigh number's meaning remains the same.  I think it's fair to expect someone trying to run with internal heating to ensure that they set f6 and c10 consistently.

I'm going to go ahead and merge this and let @illorenzo7 add the fix for initialize_reference_heating that he mentioned.

-Nick

PR #218, opened 2019-08-14
Ref mod2

feathern:
This PR follows onto [PR 215](https://github.com/geodynamics/Rayleigh/pull/215 ).   It prevents rayleigh functions and constants from being overwritten by those read in from a custom diffusivity file used in tandem with one of the Rayleigh-provided reference states.  Only those constants/functions pertaining the a custom diffusivity (specified via nu{kappa,eta}_type =3) are now overwritten by the contents of the coefficient file.  Prior to this PR, anything in the coefficient file (say density) would overwrite values previously set through initialize_reference_state.


orvedahl:
This looks good to me, approving.

PR #219, opened 2019-08-14
Set heating coefs

illorenzo7:
This is the fix I mentioned to set c_10 and f_6. I believe it should work in all cases. f_6 should always volume-integrate to 1, leaving c_10 to be the heating_integral or luminosity set in main_input. 

This pull request has three commits to make it modular in case I screwed something up. 

1. The first commit gets rid of the setting of c_10 and f_6 in the subroutine Constant_Reference(), as @orvedahl, @feathern, and I discussed. 

2. The second commit sets c_10 and f_6 in Initialize_Reference_Heating()--this should work for all cases, but please comment if I am mistaken. 

3. The third commit gets rid of what appears to be a no longer used flag: adjust_reference_heating. I think this flag used to normalize the heating based on the desired luminosity AND whatever energy is lost (or possibly pumped in?) through the boundaries by conduction if fixed-flux boundary conditions were set. I personally think it's good the code no longer does this, since now the user is required to know what they are doing when setting fixed-flux conditions coupled with internal heating. But if the logic was removed by mistake somewhere along the line, I figured I should call attention to it. It's a separate commit so can easily be removed from this pull request if need be.

### DIFF
--- a/src/Physics/BoundaryConditions.F90
+++ b/src/Physics/BoundaryConditions.F90
@@ -271,6 +271,9 @@ Contains
             Endif
 
             ref%heating = ref%heating*(lum_top-lum_bottom)
+            ! If the code has reached this point, c_10 was not defined in
+            ! initialize_reference_heating and needs to be defined here
+            ra_constants(10) = lum_top - lum_bottom
             !If something has been set inconsistenly, this will result
             ! in zero reference heating
          Endif
--- a/src/Physics/PDE_Coefficients.F90
+++ b/src/Physics/PDE_Coefficients.F90
@@ -1207,7 +1207,8 @@ Contains
         ! read_custom_reference_file()
         ! For reference_type = 1, the coefficients have been set in Constant_Reference()
         ! In all cases, f_3, f_5, f_7, c_5, c_6, c_7 have already been set in 
-        ! Initialize_Diffusivity()
+        ! Initialize_Diffusivity() and c_10 and f_6 have been set in
+        ! Initialize_Reference_Heating()
         If (reference_type .eq. 2 .or. reference_type .eq. 3) Then
             ra_constants(1) = ref%coriolis_coeff
             ra_constants(2) = 1.0d0 ! buoyancy coefficient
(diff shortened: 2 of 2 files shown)


orvedahl:
Currently, the ra_function is set around line 586 and is based off ref%heating. Then there is an if statement that can modify ref%heating. I think the ra_function should be set at the very end of the subroutine, so any modifications to ref%heating make it into the f_6

orvedahl:
Also, I think Nick means f_6 = ref%heating * rho * Temp. The constant c_10 should not be included.

illorenzo7:
Sorry, I'd forgotten about the factors of rho and T. The latest commit should fix things. I agree with Ryan's statement f_6 = ref%heating*rho*T, as long as this happens _before_ the heating is renormalized. Otherwise I will have to include the factor of c_10 in the denominator. 

I am realizing I didn't take into account that BoundaryConditions can modify the heating further if the heating_integral is > 0 and the user has specified fixed-flux boundary conditions. @feathern  Is the routine Transport_Dependencies() still working correctly? It looks like ref%heating is getting multiplied a second time a factor of luminosity for both boundaries fixed-flux. And then for one boundary fixed flux and the other fixed temperature (or entropy), the heating is getting set to zero, which isn't good either. If transport_dependencies _is_ still correct, then it looks like c_10 needs to be modified here as well, as in

ra_constants(10) = ra_constants(10)*(lum_top - lum_bottom) 

around line 273 of BoundaryConditions().

illorenzo7:
Sorry, I meant Transport_Dependencies() will set the heating to zero if, e.g., the bottom boundary is fixed flux (with zero incoming flux), and the top is fixed temperature.

illorenzo7:
Okay, I spoke with @feathern and now I understand the purpose of adjust_reference_heating (i.e., to scale the heating to the luminosity driven through by the conductive flux, if the user doesn't specify luminosity or heating_integral in main_input. I just made a commit to set c_10 in BoundaryConditions if adjust_reference_heating is .true.; the definition makes c_10 still the luminosity, effectively, with f_6 a radial "shape" integrating to 1.

feathern:
Let us now move on with our lives :)

PR #220, opened 2019-08-16
add option to cause orderly shutdown by touching a file

tukss:
This makes it possible to cause Rayleigh to exit and write a checkpoint
file by touching a file called "terminate" in the run directory. This is
convenient for restarting with new options without losing any progress.
With the right logic in the job script, even the current allocation in
the queue can be kept.

As checks for the existence of a file can be costly on some parallel
file systems, the default is to check only every 50 iterations. This can
be configured with `terminate_check_interval`.

The section "Running the code" in the documentation has been updated to
describe this feature.

### DIFF
--- a/doc/source/User_Guide/running.rst
+++ b/doc/source/User_Guide/running.rst
@@ -209,6 +209,17 @@ will run for is determined by the value of the namelist
 *max_time_minutes minutes* or when it has run for *max_iterations time
 steps* – whichever occurs first.
 
+An orderly shutdown of Rayleigh can be manually triggered by creating a file
+with the name set in **terminate_file** (i.e., running the command *touch
+terminate* in the default setting). If the file is found, Rayleigh will stop
+after the next time step and write a checkpoint file. The existence of
+**terminate_file** is checked every **terminate_check_interval** iterations.
+The check can be switched off completely by setting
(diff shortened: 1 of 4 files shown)


feathern:
Hey Philipp,

This is great.  I have one small requested modification.   Performing exact comparisons on noninteger numbers with .eq. always makes me nervous.  I'm not saying there's not somewhere else in the code where I did it myself, but  can you possible give it the check on the terminate signal's value a small range around 1?  Other than that, looks good.
-Nick

tukss:
> This is great. I have one small requested modification. Performing exact comparisons on noninteger numbers with .eq. always makes me nervous. I'm not saying there's not somewhere else in the code where I did it myself, but can you possibly check whether the terminate signal's value is within a small range around 1 -- rather than .ne. 0? Other than that, looks good.

I am a bit uneasy about using reals for this purpose anyway, but I see why we want to broadcast just one array of one type. Usually 0 (or any integer smaller than the size of the mantissa of 56 bit) should be safe, but I agree we should be careful in case of overzealous compilers. I'll implement it with a range instead.

tukss:
The new version should be safe as long as `0.0d0 < 0.5d0 < 1.0d0` holds.

PR #221, opened 2019-08-16
output iter/sec on every iteration

tukss:
This allows us to monitor the code performance during a run. It also
includes iter/(sec*ncpu), which is a basic indicator of scaling
efficiency when changing the number of processes.

The new status line looks like this:
```
  On iteration : 00000004    DeltaT :   1.0000E+03            0.23 iter/sec,       0.12 iter/(sec*ncpu)
```

I normally use sec/iter instead of iter/sec, but I picked the later to be consistent with the timing output when the code terminates.

This uses an existing timer (`walltime`), so it should not incur any significant overhead.

As this is quite a prominent change in the log output, I think we should collect a few opinions before merging.

### DIFF
--- a/doc/source/User_Guide/io_redirection.rst
+++ b/doc/source/User_Guide/io_redirection.rst
@@ -43,3 +43,8 @@ force a text-buffer flush. Unexpected crashes and sudden termination by
 the system job scheduler do not force a buffer flush. Note that the
 default value of stdout_file is **‘nofile’**. If this value is
 specified, output will directed to normal stdout.
+
+To save on disk space for logs of very long runs, the number of status outputs
+can be reduced by specifying **statusline_interval** in the
+**io_controls_namelist**. This causes only every n-th status line to be
+written.
--- a/doc/source/User_Guide/running.rst
+++ b/doc/source/User_Guide/running.rst
@@ -215,10 +215,11 @@ terminate* in the default setting). If the file is found, Rayleigh will stop
 after the next time step and write a checkpoint file. The existence of
 **terminate_file** is checked every **terminate_check_interval** iterations.
 The check can be switched off completely by setting
-**terminate_check_interval** to -1. With the appropriate job script this
-feature can be used to easily restart the code with new settings without losing
-the current allocation in the queuing system. A **terminate_file** left over
-from a previous run is automatically deleted when the code starts.
+**terminate_check_interval** to -1. Both of these options are set in the
+**io_controls_namelist**. With the appropriate job script this feature can be
+used to easily restart the code with new settings without losing the current
+allocation in the queuing system. A **terminate_file** left over from
+a previous run is automatically deleted when the code starts.
 
 Time-step size in Rayleigh is controlled by the Courant-Friedrichs-Lewy
 condition (CFL; as determined by the fluid velocity and Alfvén speed). A
--- a/src/Physics/Controls.F90
+++ b/src/Physics/Controls.F90
@@ -108,11 +108,13 @@ Module Controls
     ! What is normally sent to standard out can, if desired, be sent to a file instead
     Integer :: stdout_flush_interval = 50  ! Lines stored before stdout buffer is flushed to stdout_unit
     Integer :: terminate_check_interval = 50  ! check for presence of terminate_file every n-th time step
+    Integer :: statusline_interval = 1  ! output status information only every n-th time step
     Character*120 :: stdout_file = 'nofile'
     Character*120 :: jobinfo_file = 'jobinfo.txt'
     Character*120 :: terminate_file = 'terminate'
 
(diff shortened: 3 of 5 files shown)


illorenzo7:
As long as it really doesn't increase the simulation time, I'd be good with this. Currently it's a bit hard to get timing info after the fact (if, for example, the job is killed early) and this would be a good workaround. 

I am in approval!

One (perhaps tangential) question: how hard would it be have the logfile also output the timing info when it creates an output file? (Thinking this would be a separate pull request). I have always wondered how long this takes. It would also be useful since I have found that sometimes the MPI on Pleaides gets substantially deadlocked on output, but it's hard to know the scope of the problem without a quantitative measure of how long output takes. 

Regarding file sizes, I have found the current logfiles take up max ~1 GB for a very long run (20M iterations). "less" can still navigate these files adequately however, with perhaps a 10-15s overhead on "calculating line numbers" and searches. I don't imagine adding another column will substantially change this. If we do want the files to be smaller, we might arrange a way to have information be printed every nth timestep instead of each timestep. But that can all happen later.

feathern:
I would leave this as-is for now.  I'm thinking that having a separate timing output might be the ultimate way to go.  That way, the timing info is always accessible, even if the user didn't redirect output to a specific file and the HPC system is hanging onto the output until the job is done.  CU Summit doesn't do that, but Pleiades does.

I agree - having some measure of I/O time will be useful.
-Nick

tukss:
In commit d5edeef I added a small change in the way we write the buffered logs. This way `ifort` won't insert line breaks where it thinks we want them. It has no effect on the output in `gfortran`.

tukss:
> Regarding file sizes, I have found the current logfiles take up max ~1 GB for a very long run (20M iterations). "less" can still navigate these files adequately however, with perhaps a 10-15s overhead on "calculating line numbers" and searches.

I know the issue. You can use `less -n` to switch off line number counting (or just press `Ctrl-C` when it's counting).
1 GB log files sound like a waste of disk space obviously. They should be a good target for `gzip` and then reading them with `zless`.

>I don't imagine adding another column will substantially change this. If we do want the files to be smaller, we might arrange a way to have information be printed every nth timestep instead of each timestep. But that can all happen later.

I agree with that and might implement that later today.

tukss:
> I agree with that and might implement that later today.

I implemented that in 5c03603. The default is still to output every status line.

feathern:
@tukss   This looks pretty good to me, including statusline_interval update.   That for going ahead and documenting that! 
 
I finally had a chance to run this version and look look at the output.  I have a couple of suggestions:
1.)  The iter/(sec*cpu) is useful, but it's also redundant information since ncpu is printing in the preamble of the log file.   I think we should probably remove that in case someone wants to add some other info onto the status line later.

2.)  The formatting is a bit inconsistent for the iter/sec output.   Looking at it now, I think we should adopt the same formatting at that used for DeltaT.  This is both in terms of using the dtfmt format string, and also in terms of having the output look like "iter/sec : 1.0000E=4" as opposed to "1.0000E-4 iter/sec,"     -- That way things stay consistent. 

Also, I was able to get about 1,000 time steps per second by running a really tiny case.  The present formatting produces some funky output once you reach 100 iter/sec (which happens in production mode on Pleiades all the time).  What I see is:
`On iteration : 00000001    DeltaT :   1.0000E-04            -1.0 iter/sec,       -1.0 iter/(sec*ncpu) ` 

`On iteration : 00000002    DeltaT :   1.0000E-04             70. iter/sec,        4.4 iter/(sec*ncpu) ` 

`On iteration : 00000003    DeltaT :   1.0000E-04         0.98E+3 iter/sec,        61. iter/(sec*ncpu) `

3.)  Finally, I'd feel  more comfortable if the initial iter/sec were zero instead of -1.

Ultimately, this is what I think would look best + leave some room for additional info on the status line: 
`On iteration : 00000001    DeltaT :   1.0000E-04    Iter/sec : 0.0000E0 ` 

`On iteration : 00000002    DeltaT :   1.0000E-04    Iter/sec : 7.0000E1 ` 

`On iteration : 00000003    DeltaT :   1.0000E-04    Iter/sec : 9.800E+2 `

That's a little bit crowded, but I can fix the spacing in a quick followup pull request of my own.

Thanks!
-Nick

tukss:
I tried to implement all your suggestions. This is what the new status line looks like:
```
 On iteration : 00000002 DeltaT :  1.0000E+03 Iter/sec :  1.483E-1
```
I use a different format string for the Iter/sec item to limit the exponent to 1 digit. I guess more than 1e10 and less than 1e-10 Iter/sec are not common.

I tried to keep the padding space efficient. The Intel compiler throws a warning if we don't leave a space for a possible negative sign, so I kept it even though this shouldn't happen.

feathern:
Thanks Philipp,
I'll do a quick followup PR this week and try to space things out a bit more -- I see their pretty crammed.  Merging this for now.
-Nick

PR #225, opened 2019-09-10
I/O Format Control

feathern:
This commit adds the ability to control (a) the number of digits in Rayleigh's integer output and (b) the number of digits to the right of the decimal place in scientific-notation output.   Three variables have been added to the io_controls_namelist to enable this change:
1.) integer_output_digits (number of 0-padded digits for integer output; default 8)
2.) integer_input_digits (same but for integer input -- in case reading a Checkpoint with fewer/more digits; default 8)
3.) decimal_places (specifies number of digits to right of decimal in sci-not output; default 3)

I've also added documentation for using these new features and modified all diagnostic, checkpoint and statusline output appropriately.

Oh, also, I noticed that all files in the Utility directory were I/O related and so moved them to the IO directory.

Maybe @cianwilson wants to have a look at this?
-Nick


feathern:
Anyone else able to have a look at this?  I'd like to get this wrapped into the code soon.

orvedahl:
I can take a look at this later tonight.

PR #227, opened 2019-09-25
Replace MPI_TYPE_EXTENT with MPI_TYPE_GET_EXTENT

cianwilson:
MPI_TYPE_EXTENT has been deprecated for a while apparently so this pr gets rid of it.

I've tested this for backwards compatibility to 2.1.1 and compiled it against 4.0.1.


tukss:
Looks good to me. I doubt we have any MPI-1 only installs out there anymore.

feathern:
Just a heads up -- I plan to take a look at this sometime later this week.  Things got busy last week and this slipped by the wayside.   Thanks for adjusting this Cian!
-Nick

PR #228, opened 2019-10-01
do not link against LAPACK if using OPENBLAS

tukss:
`libopenblas` already comes with a full set of LAPACK routines included. While you can link another LAPACK library at the same time, this might override the (possibly) more optimized functions from OPENBLAS. Also this way Rayleigh doesn't compile on systems that only provide `libopenblas` but not `liblapack`.

This patch avoids linking against `liblapack` if the `--openblas` option is set.

OPENBLAS can be compiled in a non-default mode that does not include LAPACK. I have never seen this on any production system. If this is a concern, we could add an additional option or maybe even an automatic detection.

### DIFF
--- a/configure
+++ b/configure
@@ -776,7 +776,7 @@ else
     #BLAS
     if [[ $USEOPENBLAS == "TRUE" ]]
     then
-	# OPENBLAS already comes with LAPACK
+    # OPENBLAS already comes with LAPACK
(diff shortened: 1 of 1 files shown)


feathern:
Just a heads up -- if someone else wants to review this PR, please don't wait on me.  Will be a bit longer before I have time to have a look.

PR #230, opened 2019-10-24
Dipole Field Boundary Conditions

feathern:
This pull requests fixes one bug and adds one feature:
1.) The value of br_bottom set when  impose_dipole_field = .true. was not reflected in the dipole initialization routine.  Moreover, the coefficient for C(ell=1, m=1) was wrong by a factor of sqrt(2), causing a tilted dipole to have a lower amplitude than one purely aligned with the z-axes.  These inconsistencies have been fixed.

2.)  A dipole naturally satisfies the potential field boundary condition at the upper boundary.  This means there is no inconsistency between using potential match at the top and imposed dipole field at the lower boundary.   This can now be done by setting dipole_field_bottom = .true. in boundary conditions namelist.

### DIFF
--- a/src/Physics/BoundaryConditions.F90
+++ b/src/Physics/BoundaryConditions.F90
@@ -42,6 +42,7 @@ Module BoundaryConditions
     Logical :: Fix_poloidalfield_top = .False.
     Logical :: Fix_poloidalfield_bottom = .False.
     Logical :: Impose_Dipole_Field = .False.
+    Logical :: Dipole_Field_Bottom = .False.
 
     Real*8  :: T_Bottom     = 1.0d0
     Real*8  :: T_Top        = 0.0d0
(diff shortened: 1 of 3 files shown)


feathern:
Looks like one of the generic input tests uses a dipole field, comparing that generated in a python script to the one in Rayleigh.   I increased the ell =1, me = 1 coefficients by a factor of sqrt(2) as
`../../../pre_processing/rayleigh_spectral_input.py -m 1 0 2.973662220170157 -m 1 1 0.7415243282628127+0.j -o ctop_init_bc
`
`
../../../pre_processing/rayleigh_spectral_input.py -m 1 0 8.496177771914736 -m 1 1 2.1186409378937510+0.j -o cbottom_init_bc
`

I suspect I need to translate my c's to cian's c's... @cianwilson  if you have any advice let me know.  I'm going to see if I can reproduce your original c-values in the meantime.

feathern:
OK!  Looks like there is a factor of 2 difference in coefficient amplitudes specified for generic_input and those reported by Rayleigh.  This doesn't surprise me for m != 0 modes (due to the in-place transform), but it does surprise me a little for the m = 0 modes.  I'll have to look into this later.  In any event the reason for the failing tests was that I needed to update the generic inputs to reflect the new values of c1{0,1}_{top,bottom} in run_test.sh.  The ones that were there were consistent with the erroneous dipole initialization/B.C.s.   This is ready to review now.

PR #234, opened 2019-11-21
Sign error in magnetic diffusion diagnostics

orvedahl:
I think there is a sign error in how the magnetic diffusion term and the "diffusion work" terms are calculated. The induction equation has a term on the RHS that looks like "- del x [eta del x B]" and the work term is the inner product of that term with B. Using a few vector identities this term can be written as

- del x [eta del x B] = eta del^2 B + J x grad(eta) = eta del^2 B + [0, Jphi*dr(eta), -Jtheta*dr(eta)]

The sign error is in front of the "J x grad(eta)" term. Someone should definitely check my math though.

### DIFF
--- a/src/Diagnostics/Diagnostics_Induction.F90
+++ b/src/Diagnostics/Diagnostics_Induction.F90
@@ -955,7 +955,7 @@ Contains
                         2.0d0*costheta(t)*buffer(PSI,dbpdp) ) )
 
                 ! Add the contribution from a gradient in eta
-                qty(PSI) = eta(r)*(del2b-buffer(PSI,curlbphi)*dlneta(r))
+                qty(PSI) = eta(r)*(del2b+buffer(PSI,curlbphi)*dlneta(r))
 
             END_DO
 
@@ -986,7 +986,7 @@ Contains
                         ovs2theta(t)*(   buffer(PSI,bphi) - &
                         2.0d0*costheta(t)*buffer(PSI,dbtdp) ) )
 
-                qty(PSI) = eta(r)*(del2b+buffer(PSI,curlbtheta)*dlneta(r))
+                qty(PSI) = eta(r)*(del2b-buffer(PSI,curlbtheta)*dlneta(r))
 
             END_DO
 
--- a/src/Diagnostics/Diagnostics_Induction.F90
+++ b/src/Diagnostics/Diagnostics_Induction.F90
@@ -1033,7 +1033,7 @@ Contains
             Endif
             If (compute_quantity(idiff_work_pp)) Then
                 DO_PSI
-                    tmp1(PSI) = qty(PSI)*buffer(PSI,br)
+                    tmp1(PSI) = qty(PSI)*fbuffer(PSI,br)
                 END_DO
             Endif
         Endif
@@ -1062,7 +1062,7 @@ Contains
             Endif
             If (compute_quantity(idiff_work_pp)) Then
                 DO_PSI
-                    tmp1(PSI) = tmp1(PSI)+qty(PSI)*buffer(PSI,btheta)
+                    tmp1(PSI) = tmp1(PSI)+qty(PSI)*fbuffer(PSI,btheta)
                 END_DO
             Endif
         Endif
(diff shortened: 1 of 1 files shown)


orvedahl:
I should also point out that the "work" terms are missing a normalization if they are interpreted as source terms in the true magnetic energy equation. They should be multiplied by ref%Lorentz_Coeff, but since this is a constant, it can easily be handled in post processing.

I can add this in if it should be fixed, but the documentation for the quantity codes is pretty clear as to what is being calculated. Maybe it's just the source code variable names that are slightly misleading, but I would not have had a better suggestion for that.

illorenzo7:
@orvedahl, I checked your math and I think you are right. However, I am also noticing (I think) the same error occurring in the diffusion work terms that was appearing in the thermal energy for the pull request I just sent out (now approved). Namely, a confusion between buffer and fbuffer. In the idiff_work_pp terms, when the field is multiplied by the corresponding induction component (building the inner product) buffer is used instead of fbuffer. I see these errors on lines 1036, 1065, 1093 (not anywhere else, but maybe double check). Would you be able to correct these as well? Happy to approve this, since it will be more correct than it was before, but may as well fix everything now.

illorenzo7:
Or, since these are logically separate problems, you could do it as two pull requests, whatever makes sense.

illorenzo7:
@orvedahl, regarding the missing source term, I agree the names are misleading. But maybe this should happen in a separate pull request, since the documentation (in terms of math is correct). So adding the source term would also require editing the documentation to include the appropriate c (c4, I think) in front of the math. Separately, I'd also like to do something about "ecrossb"; what's being computed currently has the wrong sign, and also isn't actually "E x B" by a factor of (1/c). Currently the "ecrossb" terms are the Poynting flux (which is the quantity of interest), except they have the wrong sign and are again missing a c4. As a broader issue, can we change quantity code names/meanings, or will that just confuse everyone who is applying post-processing fixes to get what they want?

orvedahl:
I'm not sure you understood my comment about the normalization, so I'll be more explicit just to be sure: The Boussinesq induction equation (only because it has less terms) is

dt(B) = B.grad(U) - U.grad(B) + eta del^2 B

If you take the inner product of the above equation with B, the "work" terms that are currently calculated are the RHS terms and look like

B . [B.grad(U)] and B . [-U.grad(B)] and B . [eta del^2 B]

The LHS would be dt(B^2 / 2), but that is not the magnetic energy density, it is missing the normalization factor; a factor of mu_0 or a factor of 4 pi, whichever you prefer. The source code variable names are actually a good description, but the source code does not calculate the true "work" terms.

illorenzo7:
Hmm...I thought I understood but could also be confused. I agree with you that to actually get the energy density on the LHS (and thus the "true work" terms on the RHS), you need the factor of 1/mu_0 for SI units (or 1/4pi for Gaussian units) to multiply the induction equation. For the anelastic case, 1/4pi is just c_4. I think c_4 would be appropriate in all the cases. In the Boussinesq case for example, if we call the kinetic energy 1/2 f_1 v^2, then I believe we need to call the magnetic energy 1/2 c_4 B^2 in order for the total energy to be conserved. 

Sorry if I further confused things by brining up "ecrossb," that is really a separate issue. 

As it stands the documentation is consistent with specifying what's actually computed, I agree. It's just that the name "work" (in the true sense of the word) is misleading. I'd be in favor of ultimately changing what's computed (and the corresponding documentation) to be consistent with the names (and in the case of ecrossb, changing the name to "poynt_flux" or equivalent, and changing the sign of what's computed).

feathern:
Sorry for the confusion on the energy.  When I coded that up, I just forgot the normalization term.   That said, what's done is done.  This issue in my mind isn't so much one about what people would prefer vs. what's consistent with the documentation.  It's that if we change the formulas used for those outputs, people who update their code may end up having runs with inconsistent data sets.  Let's just leave it un-normalized for now.   We're getting close to having what I would consider Rayleigh 1.0.  In that version, the checkpoints will be changed as we discussed at the workshop, and so that may be a good time to revisit the issue of the quantity codes and any other related issues.

@illorenzo7  The variable names are just listed in that table so that if someone wants to grep the code for the formula, they can do so.   If the ecross_b name is bothersome to you, or you find it misleading, and you want to change it --- consistently everywhere --- that would be a harmless modification  (just don't change the actual formula associated with the quantity code).

PR #235, opened 2019-11-21
Automatic Domain Decomposition

feathern:
Firstly, please don't merge this yet!  I'm looking for feedback + a little beta testing if anyone has the time.  

This PR adds new functionality to Rayleigh.  Sometimes I get tired of specifying nprow and npcol every time I run the code.  I also suspect it's a little annoying to think about the process grid when you're just trying to run the code for the first time.   With this PR,  if nprow and npcol are not specified in main_input or at the command line, Rayleigh will identify the nprow/npcol pairs whose product equals ncpu.  It will then identify all pairs that are consistent with the chosen problem size and finally attempt to pick a sensibly optimal pair.

This way, you can just say "mpiexec -np 8 ./rayleigh.opt" and it will hopefully run.

Here's a summary of how the optimal pair is chosen.   The best-case scenario is always checked first and chosen if it exists.  Beyond that, however, this can likely be improved, so comments and suggestions are welcome.

Suitable pairs of nprow/npcol fall into distinct categories of load-balancing and process configurations.
For load-balancing, there are 4 possibilities:  
1. radius and the horizontal direction are perfectly load-balanced
2. only the horizontal direction is perfectly load-balanced
3. only the radial direction is perfectly load-balanced
4. neither direction is perfectly load-balanced

For process-grid configurations, we have:
1. nprow = npcol
2. nprow > npcol
3. nprow < npcol

That gives 12 total possibilities.   The ideal combination is perfectly load-balanced in both directions and nprow = npcol, but that's not always possible.  For each category, rayleigh identifies, if it exists, the nprow/npcol combination with the smallest difference between nprow and npcol.   Rayleigh then checks each category in a prioritized order until it finds a combination whose difference is within some (presently arbitrary) distance of the minimum across all categories.  The latter will give the best network performance latency-wise, but it could be poorly load-balanced, which is why we don't pick that one right off the bat.  nprow=npcol and perfect load-balancing are always checked first (and should be), but beyond that, I'm open to suggestions. 

I would also love someone to give it a try and see if you feel like it gives sensible results or if you observe any segmentation faults etc.

Thanks and Happy Thanksgiving!

### DIFF
--- a/src/Parallel_Framework/Parallel_Framework.F90
+++ b/src/Parallel_Framework/Parallel_Framework.F90
@@ -160,7 +160,7 @@ Contains
             Write(istr,'(i6)')self%npcol
             call stdout%print(" ---- NPCOL : "//trim(istr))
         Endif
-        If (self%nprow .le. 0) Then
+        If (self%nprow .le. 1) Then
             If (self%gcomm%rank .eq. 0) Then
                 Call stdout%print('........................................................')
                 Call stdout%print(' --- Error:  nprow must be at least 1.  Exiting...')
@@ -168,7 +168,7 @@ Contains
             Call self%exit()
         Endif
 
-        If (self%npcol .le. 0) Then
+        If (self%npcol .le. 1) Then
             If (self%gcomm%rank .eq. 0) Then
                 Call stdout%print('........................................................')
                 Call stdout%print(' --- Error:  npcol must be at least 1.  Exiting...')
--- a/src/Physics/ProblemSize.F90
+++ b/src/Physics/ProblemSize.F90
@@ -32,7 +32,7 @@ Module ProblemSize
 
     !//////////////////////////////////////////////////////////////
     ! Processor Configuration
-    Integer :: ncpu = 1, nprow = 1, npcol =1 , npout = 1
+    Integer :: ncpu = 1, nprow = -1, npcol =-1 , npout = 1
     Integer :: my_rank      ! This is the rank within a run communicator
     Integer :: global_rank  ! This differs from my_rank only when multi-run mode is active
     Integer :: ncpu_global  ! Same as ncpu unless multi-run mode is active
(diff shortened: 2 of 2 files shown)


feathern:
One more note.  I will remove those commented print statements once I have any comments/beta-testing, if any.

tukss:
This is a really useful feature. I once wrote a similar routine for the SLH code. I had a slightly different logic, but this actually looks simpler and doesn't need a routine to compute all prime factors of a number.

Is there a specific reason why we don't allow pairs where one or both factors are 1? In the current implementation the automatic domain decomposition fails for jobs with just one CPU, which are sometimes useful for debugging.

If I comment
```fortran
If (Min(f1,f2) .ge. 2)
```
everything still seems to work fine and the code accepts one CPU jobs and jobs with a prime number of CPUs.

feathern:
Good question.  There is a bug regarding serial jobs somewhere in Rayleigh, but I'm not sure where.  Even if the code runs with nprow = 1 and/or npcol = 1, you will get incorrect results when running the benchmark.  I suspect it's something to do with how I'm calling alltoallv, but I haven't really had time to look.  In any event, the current state of the code is that nprow and npcol must both be at least 2 to get accurate benchmark results consistently.

feathern:
First, @tukss   Thanks for having a look!  @ everyone else --  While I'm not in a huge hurry to merge this commit, I would like to do it before the holidays.  @orvedahl  , @cianwilson  , @illorenzo7   -- Do you all have any opinions?  You all also submit regular modifications.   I guess the main thing is that this adds additional functionality, but it's fairly maintainable, and could pretty easily be improved on if someone has an idea.   If people are OK with this, I will clean up the code a bit more and then merge, but I'm not trying to rush it.

orvedahl:
I really like this idea, but have not had a chance to run with it yet. I think what you have should be sufficient, and if it isn't then the user is probably using a really weird combination of core count and resolution, such as prime numbers. Either way, it should be relatively simple to modify/improve what you have if someone has a better idea...I'm fine with merging it as is.

tukss:
The only thing I can think of is that we should add a generic warning about using a value of `nprow` or `npcol` equal to 1 if that is an issue as you say. That's not specific to automatic domain decomposition though and could go in a separate PR.

feathern:
Thanks all.  I'm going to do some cleanup, and then approve/merge this.  If at any point, anyone cares to improve this, please do!   I think the heuristics used can be improved quite a bit, but hopefully folks running big jobs take the time to profile a little on their own!

PR #237, opened 2019-12-12
(Almost) Unified Parallel I/O

feathern:
I tag @cianwilson  @tukss  @ -- Kyle (?) @orvedahl @illorenzo7  in this one...

***This PR is fairly large, but it saves a bit over 2,000 lines of code and changes almost nothing from the user-based perspective.***  The hope is that it allows the I/O in Rayleigh to be much more maintainable than it was previously.

This PR stems from two aspects of the hackathon.   I would like to revise the Checkpointing structure, and  @cianwilson created a very nice Generic_Input interface.  

As some background, the Parallel I/O in Rayleigh was developed on-the-fly as we were in a phase of capability-demonstration for the INCITE project.  At the start of INCITE, I used serial I/O with all data gathered onto a single processes...   The result of that scramble was a jumbled mess of functional I/O-related code.

This PR provides a unified interface for all Parallel I/O (and all diagnostics) in Rayleigh, with the exception of Generic_Input.  I will touch base with @cianwilson next week for that.  In the end, all calls to MPI_IO funnel down to the same two spots (one for I, one for O) in the code.   There is a new Generic_Input_New module as part of this PR.  That is a skeleton module, but even if it's working, the relevant python routines need to be modified. Please just ignore it when reviewing this PR -- it isn't used by Rayleigh right now.  That's for @cianwilson to work from.  We'll remove the _new version once his subsequent PR is accepted.   

***This PR LARGELY affects:***  Spherical_IO.F90 (diagnostics), Parallel_IO.F90 (diagnostics, new file), and Checkpointing.F90 .  It also affects some ancillary files in a very small way as can be seen by the github diffs.

***From a user standpoint, this PR changes:***  Almost nothing.   The alternative checkpoint types, probably only used by one user, have been removed.  They can be accessed by the outputs_per_row keyword (see below).

***This  PR adds:***  outputs_per_row (io_controls_namelist; see below), caching for G_Avgs and Shell_Avgs.  Those are activated via globalavg_cache_size and shellavg_cache_size.   

***My Opinion***:  All variables should have caching enabled  (minor change), but in common practice,  G_Avgs, Shell_Avgs, and  Point_Probes are the only ones that should be cached due to memory usage.  Still, after someone's given a sense of what they think, I will add caching for the others -- they all use the same interface, and so it is trivial.

***Summary:***   All diagnostics and checkpoints have been refactored to use Parallel_IO.F90.  This new file provides a single class that handles all parallel I/O in Rayleigh.  While previously coded as serial I/O+ancillary communication, G_Avgs and Shell_Avgs have also been refactored to work within this new framework.  This way, all diagnostic outputs are programmed in a similar fashion.

All Parallel I/O is conducted from either hybrid spectral (r, ell/theta, m) or physical (r, theta, m/phi) space.  Load-balancing is such that discrete radial levels are distributed across the npcol process rows.  All parallel/diagnostic I/O now works by cascading down-row to a subset of ranks within each row.   If n_r_row is the number of radial points in a given row, then n_r_row columns within that row can be used for I/O.  

The maximum number of columns to be used is user-specified.  Use ***outputs_per_row*** in the io_controls_namelist.   If the number specified is too high, Rayleigh will user the maximum allowed number based on the number of radial points in that row and the radial indices of the particular output.  For example, some shell_slice outputs might have 3 radial indices in some rows and 1 in other rows.  If outputs_per_row is 3, the value of 1 will be used in the rows with only one radial index.  

*** My Testing:  ***   To test this, I first simply ran Rayleigh using the benchmark_diagnostics_input files (copied as main_input) from the input_examples directory.  I see nothing amiss.  Secondly, any output that is a ***sampled*** and ***not averaged*** output should give the same bit-wise result if you start from a benchmark (not random) input file.   Everything from Checkpoints through Shell_Slices seem to work fine.  If you want to test averaged outputs, the semantics are slightly different in the new version.  As a result, diff will steer you wrong due to machine-precision, and you are better off using matplotlib to plot (old_rayleigh) vs. (new_rayleigh) and convince yourself that the slope is 1.   I have done so.  I have also done so with checkpoints, and it works.  To test with caching, run with 1 ncache for the given output, then multiple, with the same executable.


feathern:
Additional Comment:  All outputs have been updated to version 5.  Prior to this PR, they were all 1-4 in their version.  No file formatting has changed, but I thought this was prudent in light of the big edit.  If something seems strange later, the versioning should help track changes back -- even if the format wasn't ***supposed*** to change.

feathern:
One more comment:  I've ran some very small tests comparing the old and new versions and haven't seen any notable differences in performance.   This makes sense because the semantics of the communication/output itself are unchanged except for  global and shell averages.

feathern:
Based on the test, looks like something's up with generic input.  I will check on that tomorrow.  I may have modified the module before the commit.  I will replace with the old module tomorrow.

feathern:
Whew!  OK.  So, I removed the generic_input_new (will send that to Cian separately to work from), and I removed index.rst, which I started to modified, but didn't finish.   There were still some testing crashes.  It appears that when you run "make fdeps," you need to do so after having run configure WITHOUT the -devel flag!   We can see about fixing this soon and/or updating to documentation so people know.   All tests have now passed.

feathern:
Thanks so much @kyle-augustson for taking time to review this hefty pull request!   I also think that more abstraction will ultimately make this useful.   Hopefully this is just a first pass to get everything into one place.   It should be a lot easier to progressively tune things now that this is done.    As far as hdf5 goes, I tend to shy away from that, but we could always make it an optional component of the build.  This is mostly me trying to avoid additional library dependencies.   Thanks again!
-Nick

PR #238, opened 2019-12-17
Main_Input Broadcast

feathern:
This PR alters how main_input is read at initialization time, with only a single rank accessing the filesystem.   Originally, main_input was read by each MPI rank.  This seemed to work, even with large jobs on Mira, but it's better practice to have one rank read main_input and broadcast its contents to the other ranks.  This keeps O(10^5) processes from trying to access the same file on disk simultaneously.   

With this PR, rank 0 (within each sub-simulation for multi-run mode) reads main_input into a character array and then broadcasts that array to every other MPI rank.   Each rank then performs a namelist read from the character array, which acts as an internal file.   If for some reason, the read fails, Rayleigh falls back onto the old method.  If that fails, an error message is displayed and the code exits with error code 13.

I've tested this with GNU and Intel compilers and things seem to work.   Comments appeared to give the Intel compiler (17.0.4) trouble, and so those are stripped out when main_input is read.


feathern:
After three weeks waiting on a review, this is finally holding up development.   Merging.

PR #239, opened 2019-12-17
Precision Module

feathern:
This PR adds a precision module to Rayleigh.   The current syntax used by Rayleigh when declaring variables is Real*8 instead of Real(Kind=X).  The former is outdated and not considered portable.   I would like to move toward a more portable format where all variables are declared with kind=x, and where x is taken from the ISO_Fortran_Env intrinsic module.

This PR is intended to lay groundwork for updating the variable declarations.  It only adds a precision module to Rayleigh that provides double-precision (ra_dp) and 4-byte-integer (ra_int) kinds.  The module Ra_Precision sits at the base of the dependency tree and is currently only "used" by Math_Constants, and only through a use statement.  Math_Constants doesn't actually use ra_dp or ra_int.   This was done so that "make fdeps" didn't run into any issues.

This should be very safe to approve/merge.

I will look into automating the changes to variable declarations in a later PR.

-Nick


feathern:
Thanks!  And I suggest we try doing it in pieces.  This might be a good way for some folks to look through the code.

PR #242, opened 2020-02-02
Portable & Complete Checkpoints

feathern:
This PR changes the checkpoint format used by Rayleigh.  It's one of the last major changes I would like to push through before a 1.0 version release (we're at 0.95 now).   These modifications are fully compatible with old checkpoint files -- nothing new needs to be done to restart from the old format.    Rayleigh checks for the new format first when reading a checkpoint.  If that's not found, it drops back to the old format.

When new checkpoints are written their format differs as follows.

1.) Each checkpoint now gets a dedicated directory into which all checkpoint files are stored.  This directory contains all information needed to restart a model.

2.) In addition to the old files (e.g., W, PAB, etc.) a main_input file, an equation coefficients file, and a boundary conditions file are written.

Since the checkpoints now have everything needed to restart a model, you can just move the relevant directory around and not worry about the main_input file, or any custom reference-state/boundary-conditions files.

In addition to the normal method of restarting, you can also restart in 'full restart' mode from the command-line using the -full_restart flag.  For example:
mpiexec -np 16 ./rayleigh.opt -full_restart 20  (to restart from iteration 20)
mpiexec -np 16 ./rayleigh.opt -full_restart -2  (to restart from quicksave 2)
mpiexec -np 16 ./rayleigh.opt -full_restart 20 -input_digits 10 (to restart from a checkpoint written with 10 digits instead of the default 8)

Main_input, etc. doesn't need to be copied to the run directory when running with full_restart.  Rayleigh gets it directly from the Checkpoint directory in this mode.  If the main_input indicates that a custom reference state or custom boundary conditions were used, those are retrieved from the relevant checkpoint files.

-Nick

### DIFF
--- a/src/IO/Parallel_IO.F90
+++ b/src/IO/Parallel_IO.F90
@@ -1,3 +1,22 @@
+!
+!  Copyright (C) 2018 by the authors of the RAYLEIGH code.
+!
+!  This file is part of RAYLEIGH.
+!
+!  RAYLEIGH is free software; you can redistribute it and/or modify
+!  it under the terms of the GNU General Public License as published by
+!  the Free Software Foundation; either version 3, or (at your option)
+!  any later version.
+!
+!  RAYLEIGH is distributed in the hope that it will be useful,
+!  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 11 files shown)


feathern:
17 days with no comments.  Merging.

PR #248, opened 2020-05-05
Fixed equation set fresh

illorenzo7:
This pull request clarifies several points in the "Equation Sets Solved by Rayleigh" section of the documentation. Specifically, it addresses the non-dimensionalization of reference_type=1,3, as well has the relationship of various functions/constants to the heating and transport coefficients. Also the documentation has been reformatted so as to not require a horizontal scroll bar.


illorenzo7:
I just checked the the equation set page on readthedocs, and it all looks fine (viewing through Chrome on Mac).

PR #256, opened 2020-08-07
Reformat vapor_notebook, fix some typos, convert Type*n to Type(kind=n) or Type(len=n)

JoelCourtney:
- reformated post_processing/vapor_notebook.ipynb to be be more readable.
- fixed a couple typos in doc/source/User_Guide/checkpointing.rst
- fixed a bug in post_processing/rayleigh_diagnostics.py that would only trigger when converting vector quantities to vapor but no scalars.


JoelCourtney:
It looks like Jenkins is telling me that I'm not allowed to changed vapor_notebook.ipynb. Is that right?

feathern:
Hey Joel,
If I recall correctly, we need to clear the cells of any notebook before pushing to the repo.  Can you try clearing all output and pushing that version to your branch?   Jenkins should automatically rerun, and then we can see.
-Nick

feathern:
I think maybe the Jenkins server is down.  I'll have a look once it's back up and the tests have run.

feathern:
Hey Joel,
Can you comment on what the bug was you were seeing?  We definitely want to keep the ofile name as 'self.tempdir/temp.cube.'  Was there a bug when rm couldn't find at cleanup time?
-Nick

JoelCourtney:
Oh, you're right. Also the indentation change I made was not necessary. I'll undo both in a sec.

The bug was: I was converting the velocity field to vapor input, but not converting any scalar quantities. Since there were no scalars being converted, ofile was never assigned because its the temp file for scalars. But the clean-up step tried to delete it anyway, so I just added the if self.nvars > 0 check first.

JoelCourtney:
My confusing over ofile's name was because I think line 1819 is redundant.

JoelCourtney:
Also I used a couple sed commands to change things like Real*8 to Real(kind=8), etc, as per https://github.com/geodynamics/Rayleigh/projects/3. I didn't mean to push that until after this PR was done, but why not.

PR #257, opened 2020-08-12
typos, bug fix, new vapor converter notebook

JoelCourtney:
- fixed two small typos in checkpointing.rst
- fixed a bug in rayleigh_diagnostics.py where cleanup would fail when converting only vector quantities to vapor input.
- added a more usable but less detailed version of vapor_notebook.ipynb called vapor_converter.ipynb.


feathern:
Thanks Joel -- looks good.

PR #258, opened 2020-09-02
Made luminosity or heating_integral specified in main_input override …

illorenzo7:
…c_10 within custom_reference framework. This shouldn't change any of the existing custom reference interface, it will just allow luminosity and heating_integral to be observed in main_input, rather than

override_constant(10) = .true.
ra_constants(10) = [luminosity]

Or else setting c_10 in the custom reference file.


feathern (approved):
Thanks Loren -- looks good.

PR #259, opened 2020-09-02
Removed duplicate error messages for transport coefficients functions/constants

illorenzo7:
These were always thrown when reference_type = 4 regardless of the diffusion_types. I also standardized the formatting of the error messages. This branch is based off the luminosity_flag branch.

### DIFF
--- a/src/Physics/PDE_Coefficients.F90
+++ b/src/Physics/PDE_Coefficients.F90
@@ -708,7 +708,7 @@ Contains
             If (ra_function_set(fi) .eq. 0) Then
                 If (my_rank .eq. 0) Then
                     Write(ind, '(I2)') fi
-                    Call stdout%print('ERROR: function f_'//Trim(Adjustl(ind))//' must be set in the custom reference file')
+                    Call stdout%print('ERROR: function f_'//Adjustl(ind)//' must be set in the custom reference file')
                 Endif
             Endif
         Enddo
@@ -788,6 +788,9 @@ Contains
         Real*8  :: input_constants(1:n_ra_constants)
         Real*8, Allocatable :: ref_arr_old(:,:), rtmp(:), rtmp2(:)
         Real*8, Allocatable :: old_radius(:)
+        Character*12 :: dstring
+        Character*8 :: dofmt = '(ES12.5)'
+        Character(len=2) :: ind
 
         cset(:) = 0
         input_constants(:) = 0.0d0
@@ -848,7 +851,10 @@ Contains
             ! Print the values of the constants
             Do k = 1, n_ra_constants
                 If (my_rank .eq. 0) Then
-                    Write(6,*)'c: ', k, ra_constants(k)
+                    Write(ind, '(I2)') k
+                    Write(dstring,dofmt) ra_constants(k)
+                    Call stdout%print('c_'//Adjustl(ind)//' = '//Trim(dstring))
+                    !Write(6,*)'c: ', k, ra_constants(k)
                 Endif
             Enddo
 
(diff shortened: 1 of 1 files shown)


feathern:
Hi Loren,
Sorry for the delay.   I'm a little confused on this one.  You avoid functions 3,5, and 7 in your fi_to_check.  That's nu, kappa, and eta, and they definitely need to be set.  You also added 14, which is dsdr, but since advect reference_state is false by default, dsdr should not need to be specified.  I would suggest addressing that last part in a separate pr (where the value of advect reference state is changed to true, while dsdr is set to 0 unless read in from a custom file).  I'm still confused on the first part, though.
-Nick

feathern:
Ah actually, I'm tracking a bit better now.  So I think we may want to add some additional logic so that an error message is thrown if those functions are not set AND nu_type etc.  equals 3.  Do you agree?

illorenzo7:
Hey Nick, thanks for checking this out! From the logic in Initialize_Diffusivity (select case 3) I believe an error is thrown for the case diffusion_type = 3 but the corresponding function is not set (line 1199, previously 1187). I will run the code really quick on Pleiades to verify that this happens in practice. 

As for dsdr, I agree that should be checked (and an error thrown) only if advect_reference_state is .true. I'll remove "14" from the fi_to_check array and update the code accordingly in a separate pull request.

illorenzo7:
Hey Nick, thanks for checking this out! From the logic in Initialize_Diffusivity (select case 3) I believe an error is thrown for the case diffusion_type = 3 but the corresponding function is not set (line 1199, previously 1187). I will run the code really quick on Pleiades to verify that this happens in practice. 

As for dsdr, I agree that should be checked (and an error thrown) only if advect_reference_state is .true. I'll remove "14" from the fi_to_check array and update the code accordingly in a separate pull request.

illorenzo7:
Hey Nick, thanks for checking this out! From the logic in Initialize_Diffusivity (select case 3) I believe an error is thrown for the case diffusion_type = 3 but the corresponding function is not set (line 1199, previously 1187). I will run the code really quick on Pleiades to verify that this happens in practice. 

As for dsdr, I agree that should be checked (and an error thrown) only if advect_reference_state is .true. I'll remove "14" from the fi_to_check array and update the code accordingly in a separate pull request.

illorenzo7:
Hey Nick, thanks for checking this out! From the logic in Initialize_Diffusivity (select case 3) I believe an error is thrown for the case diffusion_type = 3 but the corresponding function is not set (line 1199, previously 1187). I will run the code really quick on Pleiades to verify that this happens in practice. 

As for dsdr, I agree that should be checked (and an error thrown) only if advect_reference_state is .true. I'll remove "14" from the fi_to_check array and update the code accordingly in a separate pull request.

illorenzo7:
Hey Nick, thanks for checking this out! From the logic in Initialize_Diffusivity (select case 3) I believe an error is thrown for the case diffusion_type = 3 but the corresponding function is not set (line 1199, previously 1187). I will run the code really quick on Pleiades to verify that this happens in practice. 

As for dsdr, I agree that should be checked (and an error thrown) only if advect_reference_state is .true. I'll remove "14" from the fi_to_check array and update the code accordingly in a separate pull request.

illorenzo7:
Crap, sorry for the duplicate comments. My computer froze and I kept clicking the comment button thinking it wasn't doing anything :-/

illorenzo7:
OK, it looks like the errors are thrown properly when nu_type = 3 but f_3 is not set in the custom reference file. I also removed f_14 from the list of functions to set if they were set, and made some minor formatting changes so the c's get output as

c_1, c_2, ... c_10, etc.

feathern (approved):
Ah OK, I missed the error in initialize_diffusivities.   This looks good to me now.

PR #261, opened 2020-10-09
Checkpoint Size Bug Fix

feathern:
I thought I had already merged this into the main branch...

This fixes an integer-arithmetic-related bug where checkpoint files larger than 2 GB were erroneously found to have the wrong filesize at init time.  All variables used in the size-check calculation are now 8-byte Integers.


feathern:
This is now preventing other users besides myself from running large simulations.  I'm going to go ahead and merge without review.

PR #265, opened 2020-12-21
Fix misuse of angular velocity in diagnostics

illorenzo7:
Four of four. "angular_velocity" is only used for dimensional runs; 1811 and 1812 should have used ref%Coriolis_Coeff/2.0. Currently they will be computed incorrectly using angular_velocity = -1 (its default value) if user does not specify angular_velocity, as in a nondimensional run or custom run.


feathern (approved):
Good catches Loren.  I agree with the logic and the math.   I'll try to submit a separate commit removing the commented code that you modified so the sign issue doesn't creep back in.  Best to just have it in one spot I think.

PR #266, opened 2021-03-23
Shell Spectra Once Again Satisfy Parseval's Theorem

feathern:
This partially fixes a bug related to the power spectra normalization.  Generally, shell_spectra power should satisfy Parsevals's theorem, such that the integrated power (e.g., v* v) over ell and m is equal to the integral of v^2 over the shell in physical space.  For some reason, the m!=0 power is too high again by a factor of 2.  This was fixed in shell_spectra version 4, but seems to have crept back in in version 5.  

I will look into the Rayleigh source later to see where/why this happened again, but for the time being, this quick fix in rayleigh_diagnostics.py will correct the problem for all versions except for version 4.


orvedahl (approved):
A nice simple fix for now, I'm approving.

PR #267, opened 2021-03-23
Critical Bug Fix: Duplicate/Non-existent quantity codes no longer garble output

feathern:
This update allows Rayleigh to deal gracefully with an easy source of user error.   When the new I/O interface was written, the indexing logic for output data was changed slightly.  If duplicate and/or non-existent quantity codes were specified via main_input, an indexing variable was improperly incremented, leading to corrupted output.   This is a pretty easily-expected source of user error, and so Rayleigh should guard against this.  This update fixes the logical issue, and the output is no longer corrupted.  The new behavior is as follows:

1) If a valid quantity code is specified multiple times, it only appears once in the output file.  The file appears exactly as though no duplicates were specified in main_input.

2) If an invalid quantity code is specified, it does not appear in the output, but there is 'blank' space (i.e., zeros) in the file where this value would otherwise appear.  The quantity code is recorded as 4100 (maximum allowsed quantity-code value; currently unused).

3) If a magnetic quantity code is specified for a hydro run, the behavior is identical to (2).

We can certainly improve on this later if people have opinions.  The point of this update is to stop the output corruption.
-Nick

### DIFF
--- a/src/IO/Parallel_IO.F90
+++ b/src/IO/Parallel_IO.F90
@@ -157,6 +157,7 @@ Module Parallel_IO
         Procedure :: DeAllocate_Receive_Buffers
         Procedure :: write_data
         Procedure :: Spectral_Prep
+        Procedure :: Reset_Cache_Index
 
         Procedure :: gather_data
         Procedure :: cascade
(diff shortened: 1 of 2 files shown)


orvedahl (approved):
This looks good to me, I'm approving it.

PR #268, opened 2021-05-20
update Conda environment documentation

tukss:
This changes the installed packages to ensure `mpif90` and `mpicc` are available.

It also changes the way `MKLROOT` is set. `CONDA_PREFIX` is a more reliable way to find where the current conda environment is installed, particularly on macOS.

I also took the liberty to quote all the used shell variables in case someone is daring enough to have a space in their conda path.


feathern:
This looks good to me.  Thanks Phillipp!

PR #269, opened 2021-05-24
Add the bibtex files to conf.py for make doc

cianwilson:
Just add a list of the bibfiles to `conf.py`

This appears to work with sphinxcontrib-bibtex==2.2.1 but is untested on other versions.


feathern (approved):
Thanks Cian!

PR #273, opened 2021-05-24
Updated configure file for mac: lstdc++ to lc++

ksoderlu:
To correct the following error when compiling using a Mac:
ld: library not found for -lstdc++


feathern (approved):
This looks good to me.  Hopefully this works across Macs.  We can test throughout the workshop.  Thanks!

PR #274, opened 2021-05-25
a new Python interface for outputs

tukss:
This implements a new, more object-oriented Python interface to the outputs. Arrays are loaded using memory mapping if possible.

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -1,0 +1,12 @@
+name: Lint
+
+on: [push, pull_request]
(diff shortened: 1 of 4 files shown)


tukss:
Just giving an update here. The new interface is basically done and I'm quite happy with both usability and performance. I still have to write documentation. Once that's done, I'll ask for a review.

feathern (approved):
Looks good.  I look forward to seeing your demo this morning.
-Nick

PR #275, opened 2021-05-25
Fix Github Action

gassmoeller:
Do not merge yet, I need to make sure the action succeeds and can be merged.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -6,22 +6,16 @@ jobs:
   test:
     name: test-build
     runs-on: [ubuntu-latest]
-    container: gassmoeller/rayleigh-buildenv-focal
+    container: geodynamics/rayleigh-buildenv-bionic
 
     steps:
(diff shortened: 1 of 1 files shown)


gassmoeller:
Ok, this is ready for review and merge. The github actions now succeeds (https://github.com/gassmoeller/Rayleigh/actions/runs/876124464). 
This currently has two drawbacks compared to the (still active) Jenkins tester:
- I had to use openmpi, because mpich does not allow oversubscribing as easily. But mpirun creates some weird error messages at runtime inside the container (see here: https://github.com/gassmoeller/Rayleigh/runs/2668779073?check_suite_focus=true#step:6:56). The test results are unchanged though.
- I had to disable the input test for now, because I do not want to mess up the mpirun commands in the script with openmpi specific options. Maybe we can find a better way once I have the conda container build and set up. But by merging this we at least do not see failing tests for all PRs.

tukss (approved):
Approved after in person discussion

PR #276, opened 2021-05-25
Use intel option -heap-arrays to avoid segfaults when copying large arrays

cianwilson:
**Updated report:**
For larger values of lmax the array copy at Parallel_IO.F90:1173 was causing segfaults when compiled with the intel compiler.  This was originally fixed by unrolling the loop but is now fixed in this pr by using the compiler flag `-heap-arrays` set to mimix the default behavior of the gnu compiler by setting a size of 65kb at which intel will which from using the stack to using heap memory.

**Original report:**

We've occasionally hit a segfault with the intel oneapi compiler caused by this copy.  Explicitly unrolling the loop seems to fix it.

I couldn't find anything wrong with the line (arrays appear correctly sized and initialized etc.).  The segfault doesn't occur with other compilers.

Any suggestions for me to track down the problem to save us having to use a local hacked version?


gassmoeller:
I know little to nothing about how the Intel fortran compiler does these things. Is there anything helpful here: https://software.intel.com/content/www/us/en/develop/articles/determining-root-cause-of-sigsegv-or-sigbus-errors.html? 

Their first suggestion is: 
Possible Cause #1 Fortran Specific Stackspace Exhaustion. Solution: -heap-arrays compiler option.
The Intel Fortran Compiler uses stack space to allocate a number of temporary or intermediate copies of array data. 

How large is this array you are assigning?

cianwilson:
So `-heap-arrays` made it work! Thanks @gassmoeller!  Nice find.

In the example I have segfaulting the array appears to be:
`(lmax+1)*(lmax+1)*nr_local*2 = 240*240*45*2 = 5184000`

Has anyone else hit this with intel?  I'll look into some of the other solutions offered on that webpage to see if this is somehow related to our cluster setup.

cianwilson:
Increasing my local (user) stack size limit also seems to work for my example failure (without needing `-heap-arrays`):
`ulimit -s 65533`

From discussion with @gassmoeller off github it's worth noting that `-heap-arrays` does take an optional size argument at which intel switches from using the stack to using the heap.  Sounds like with the right number this might emulate the gnu compiler's behavior if that's desired/other people hit this issue.

gassmoeller:
Also from a discussion off github maybe we can figure out why the Intel compiler is creating the stack copy in the first place, both arrays should be on the heap, so it looks like an internal copy of some sort. Maybe writing the copy as `A = B` instead of `A[:,:,:,:] = B[:,:,:,:]` would fix this?

gassmoeller:
This looks good to me, and I would like to have this in the 1.0 release, because it fixes a weird crash. @tukss: Let me know how your testing goes, I would like to tag a release version late in the day today.

tukss (approved):
The new option works fine on ifort 19.0.2.187 in the sense that the code compiles and runs. I couldn't reproduce the previous errors easily, but I'm quite confident that this doesn't break things.

PR #278, opened 2021-05-25
add instructions for building and using the Docker environment

tukss:
These are some instructions for using the Docker container as a development environment. It uses our current Docker container on Docker Hub at the moment. In the future we might want to switch this to a newer build environment.

I chose to mount the home directory at `/root`. This has the advantage that `cd` goes to the right home directory and we can use the SSH keys from the normal home directory for git. I don't see a big problem in running the code as root inside the container, but if others prefer we could create an unprivileged  user inside the container. The downside would be that the users can't use `apt` directly to get new software.

### DIFF
--- a/doc/source/User_Guide/devel_environment.rst
+++ b/doc/source/User_Guide/devel_environment.rst
@@ -4,7 +4,12 @@ Setting up a Rayleigh Development Environment
 When running Rayleigh on HPC resources, always compile the software with the recommended compiler and link against
 libraries optimized for the architecture you are running on.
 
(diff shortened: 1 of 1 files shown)


doc/source/User_Guide/devel_environment.rst
```diff
@@ -113,4 +118,48 @@ At this point, you can run "make install," and run the code using mpirun as you
...
+.. code-block:: bash
+
+    cd /root/path/to/Rayleigh
+    ./configure --with-fftw=/usr
+    make
+
+
+Building the container
+^^^^^^^^^^^^^^^^^^^^^^
+.. note:: This step purely optional. You only need to do this if you cannot pull the container from Docker Hub or you want to modify the Dockerfile.
```
gassmoeller, line 159: Maybe include that this has to happen *outside* of the container, and that you need to be in the Rayleigh directory.

gassmoeller (commented):
I like it! I have one small questions, but everything else looks good.

doc/source/User_Guide/devel_environment.rst
tukss: Good point. I added a commit with more details on building the container.

feathern (approved):
Thanks!

PR #279, opened 2021-05-25
Add Github action that installs Rayleigh using conda

gassmoeller:
This adds an automatic test to every PR to build Rayleigh using conda, following the instructions here: https://rayleigh-documentation.readthedocs.io/en/latest/doc/source/User_Guide/devel_environment.html?highlight=conda#. This job and the existing job (that builds inside a docker container) run in parallel and take about the same time, so this should not slow down testing. Unfortunately I can not run the tests in the conda environment as well, because mpich does not allow oversubscription and the github runner only has 2 cores. So I only test compilation, still might be handy to have one job using lapack/blas/openmpi, and the new one using MKL/mpich.

Moreover, this PR adds an `environment.yml` file that allows to generate the Rayleigh development environment as easy as running `conda env create -f environment.yml`, which will create a ready to use `radev` environment. We may want to update the instructions on the website to use this instead of the current instructions if you think it is useful.

Lastly it uses the github concurrency feature (https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency) to make sure only a single workflow is running at the same time for each branch. I.e. if you update your pull request with a new commit, and the old github workflow is still running that one is cancelled and you new job starts. This saves some resources.


tukss:
That looks good to me and the new action for the conda environment seems to run fine.

A general question regarding our choice of conda packages. MKL comes with its own version of FFTW, so is there a reason we install it in addition to MKL? Or is that just historic?

tukss (approved):

gassmoeller:
>  so is there a reason we install it in addition to MKL? Or is that just historic?

I dont't know, let's find out what happens without it. I restarted the test without.

gassmoeller:
Without fftw in the environment compilation breaks with the following error:

```
#? Fourier_Transform.F90:23: Error: Can't open included file 'fftw3.f'
```

Let's continue with this for now and if we want to work around this by using whatever is inside MKL then let's make it a separate PR.

tukss:
Yes. I'll look into fixing the include path for MKL later. The include files should hopefully be compatible.

gassmoeller:
Ok, then I'll go ahead and merge this for now. I will do some more testing with these actions.

PR #281, opened 2021-05-26
Remove cmake build system

gassmoeller:
The cmake build system is not used, not documented, and out of date. Because no one is interested in maintaining it we should remove it for now. If it is needed again in the future it can be recovered from the git history.


orvedahl (approved):
This looks good to me, thanks Rene.

bangerth:
What a shame :-)

Related to #60.

gassmoeller:
I agree, but in the interested of not keeping broken things in the repo it seemed the best solution.

bangerth:
Yes, absolutely!

PR #282, opened 2021-05-26
Improve test action

gassmoeller:
I added an environment variable to add OpenMPI specific options into the generic_input test run script. This should avoid conflicts with running the script under different  MPI implementations.

With these changes the new github test action now runs all of the tests that the old Jenkinsfile has. In principle the Jenkins tester is now not longer needed (but we can run it in parallel for a while, before removing the Jenkinsfile).

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -37,9 +37,9 @@ jobs:
 
         mpirun -np 4 $RAYLEIGH_TEST_MPI_PARAMS ../../bin/rayleigh.dbg
 
-        // Generic input test
+        # Generic input test
         cd "$GITHUB_WORKSPACE"
-        source tests/generic_input/run_test.sh
+        sh ./tests/generic_input/run_test.sh
 
         git diff > changes.diff
         git diff --exit-code --name-only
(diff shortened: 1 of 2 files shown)


tukss:
That's a good solution.

Is there anything that Jenkins can do that the actions can't? Like running large parallel tests? Not that we do that.

tukss (approved):

gassmoeller:
> Is there anything that Jenkins can do that the actions can't? Like running large parallel tests? Not that we do that.

I do not know, there are probably a few exotic things Jenkins can do that actions cant, however I cannot think of anything we would want to use. Jenkins runs all of the tests on CIG owned hardware, which is somewhat better than the free github actions runner, but the number of parallel jobs we can run on github more than makes up for that (and if we ever need stronger hardware we can run a self-hosted runner at CIG, which is less effort than a Jenkins server). I mostly want to transition away from Jenkins because it reduces the number of things that can break or fail. Maintaining Jenkins requires continuous effort on CIG's  part, and the hardware is old and may fail soonish. Github Actions just seem like the more efficient alternative.

PR #284, opened 2021-05-26
Allow building a particular version of Rayleigh (dbg, opt, etc.)

orvedahl:
The default behavior when typing "make" is to build the debug version as well as the optimized version. This PR allows you to specify a particular version of Rayleigh to build by typing "make target=opt" or "make target=dbg". The default behavior is to build everything, just as before.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,9 +9,12 @@ export LC_COLLATE=C
 # make the CUSTOMROOT variable available to sub-make processes
 export CUSTOMROOT
 
+# running "make target=dbg" will only compile the specified target
+target=all
+
 rayleigh: prepare_directory
 	@$(MAKE) --no-print-directory --directory=$(BUILD) clean_exec
-	@$(MAKE) --no-print-directory --directory=$(BUILD) all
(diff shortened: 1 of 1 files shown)


gassmoeller (commented):
This looks good to me, I also tested the options on my system. A question about the name: Most software I know of uses "Debug" and "Release" or "debug" and "release" as the two build targets. Is there a reason you went for dbg and opt (other than the names of the binaries)?
@feathern do you have an opinion on the the names of the build targets?

orvedahl:
The src/build/Machine_Definitions file defines the different versions that will be built, stored in the VERSIONS variable. I just borrowed those names. I believe those versions are taken from the executable extension, e.g., rayleigh.opt, rayleigh.dbg, rayleigh.avx2, etc.

gassmoeller:
Ah, I see. Yes I guess in order to stay consistent we should use the same names everywhere. Let's just merge this and consider the naming scheme another time (if there is interest).

gassmoeller (approved):

Issue #144, opened 2019-07-23
'--FFLAGS' option in configure deletes preprocessor flags


hirokemono:
When configure runs with '--FFLAGS' option, preprocessor flags by configure command (such as -DINTEL_COMPILER) disappears. Here is the example,

with --FFLAGS
login4(1048)$ ./configure --with-mkl=$(TACC_MKL_DIR) --FC=mpif90 --CC=icc --FFLAGS='-O3'
-bash: TACC_MKL_DIR: command not found

 -----------------------------------------------------------------------------
                          Rayleigh v0.9 Configuration
 -----------------------------------------------------------------------------

  Overriding all flags..,
 
 Compiler manufacturer      :  INTEL
 Fortran compiler command   :  mpif90
 Fortran optimization flags :  -O3
 Fortran debugging flags    :  -O3
 Intel compiler set:  MKLROOT is defined
 C compiler command         :  icc
 
***********************************************************************
 Intel compiler detected.  Please indicate desired SIMD instruction set.

 Suggested instruction sets are: 
 -- Westmere cores                  :  SSE4.2
 -- Sandy Bridge / Ivy Bridge cores :  AVX
 -- Haswell /Broadwell              :  AVX2 
 -- SkyLake / Knights Landing       :  AVX512 
 -- Support for all of the above    :  FAT

1) SSE4.2
2) AVX
3) AVX2
4) AVX512
5) ALL
6) FAT
7) NONE
#? 6
***********************************************************************
 Include flags              :  -I/opt/intel/compilers_and_libraries_2018.2.199/linux/mkl/include -I/opt/intel/compilers_and_libraries_2018.2.199/linux/mkl/include/fftw

 Library flags              :  

 
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 Configuration is complete.
 A Machine_Definitions files has been placed in /home1/02179/matsui/Rayleigh/src/build.

 Typing 'make' will generate the following executables (optimization indicated): 
 rayleigh.opt  :   -O3 -xSSE4.2 -axAVX,CORE-AVX2,CORE-AVX512,MIC-AVX512
 rayleigh.dbg  :   -O3

==================================================================
no FFLAGS case
[...]

PR #286, opened 2021-05-27
Fix for Issue #144: internally used preprocessor flags are no longer overwritten by custom flags

orvedahl:
When running ./configure a compiler-dependent preprocessor flag is included such as "-DINTEL_COMPILER" or "-DGNU_COMPILER". When specifying custom flags such as "./configure --FFLAGS='-O3' ..." the preprocessor flags were overwritten and lost, even when using Intel and GCC compilers. This PR adds all preprocessor flags after the custom flags have been applied, i.e., fix for Issue #144

### DIFF
--- a/configure
+++ b/configure
@@ -32,12 +32,12 @@ function opt_flags {
   #$1 --version
   if [[ $1 == "INTEL" ]]
   then
-    echo '-FR -fpp -r8 -O3 -shared-intel -DINTEL_COMPILER'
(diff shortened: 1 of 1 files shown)


gassmoeller (commented):
This looks good to me, but I am not an expert on how these flags work together. @hirokemono: Please take a look at this: Does this fix your issue?

hirokemono (commented):
it looks more makes sense.

gassmoeller (approved):

Issue #260, opened 2020-09-02
Typos in rayleigh_diagnostics


janabrah:
Lines 1170-1172 in rayleigh_diagnostics (in the function Power_Spectra) read:

            br_index = a.lut[401]
            bt_index = a.lut[402]
            bp_index = a.lut[403]

I think "401,402,403" are typos, and they should be 

            br_index = a.lut[801]
            bt_index = a.lut[802]
            bp_index = a.lut[803]
            

With the Shell_Spectra output folder populated and a valid "istring", I can recreate this bug with:

    from rayleigh_diagnostics import Shell_Spectra, Power_Spectrum
    power = Power_Spectrum(istring,magnetic=True)

The error message I got was:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "rayleigh_diagnostics.py", line 1043, in __init__
    self.spectra_file_init(path+'/'+infile)
  File "rayleigh_diagnostics.py", line 1179, in spectra_file_init
    btc = np.reshape(a.vals[:,:,:, bt_index, :],dims)
IndexError: index 1000 is out of bounds for axis 3 with size 8

where the reason that failed was because a.vals was initiated incorrectly. We hadn't used the output code 402, and if we had there's a chance it wouldn't have thrown an error (our only output codes were: 1,2,3,401,801,802,803,1101). Still, even without an actual error, I think it would have been plotting things related to kinetic energy instead of magnetic field.

janabrah:
and bug report aside, thank you *so so much* for providing Rayleigh. It's absolutely been phenomenal.

feathern:
Hi @janabrah ,
Thanks for the kind words about Rayleigh.   This is indeed a bug.  If you would like, I can fix this, or alternatively, you could submit a pull request that I'll review more rapidly than I replied!
-Nick

PR #287, opened 2021-05-27
Fix Issue #260: typo in rayleigh_diagnostics.py

orvedahl:
The Power_Spectrum class referenced the wrong magnetic field quantity codes. I updated these, fixes Issue #260


gassmoeller (approved):
Looks good to me. Did you check that 801,802,803 are actually the correct quantity codes? You probably did, but I do not know the codes, so just double-checking.

orvedahl:
Yes, 801,802,803 refer to the full magnetic field components

PR #288, opened 2021-05-27
Update AUTHORS file

gassmoeller:
This removes the version number from the AUTHORS file (which would be a pain to update for every release). It also updates the text to better reflect Rayleigh's current development model. @feathern should take a look at this before merging.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,7 +1,9 @@
-Rayleigh 
-Version 0.9.0
-Developed by Nicholas Featherstone
+Rayleigh was originally developed by Nicholas Featherstone.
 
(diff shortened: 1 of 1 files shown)


ljhwang (commented):
Is the extra "and" intentional?
    **and** Ryan Orvedahl, Krista Soderlund, **and** Rakesh Yadav

If not, the first one should be removed.

orvedahl:
The extra "and" looks like a typo to me.

gassmoeller:
Fixed.

ljhwang (approved):
Thanks for the fix.

feathern (commented):
This looks good and sensible to me.

PR #289, opened 2021-05-27
Update INSTALL formatting

gassmoeller:
This updates the formatting of the INSTALL file. Two things changed:

- remove version number and author from the header, version number is irrelevant, because these instructions are for the current version (which may change), and the author is listed in the AUTHORS file.
- reformatted lines that were longer than usual line length conventions, making it easier to read in simple text editor. No text was changed, just moved line breaks to different places.


orvedahl (approved):
This looks good to me, thanks Rene.

PR #291, opened 2021-05-27
Remove duplicated version numbers

gassmoeller:
This removes some more duplicated version numbers, and reformats the VERSION file, so that it can be easily used in scripts (the only content of the file is the number now).


orvedahl (approved):
Looks good, thanks.

PR #293, opened 2021-05-27
Reduce duplication in doc and devel environments

gassmoeller:
Our development environment and the environment that builds the documentation should ideally be the same environment to make sure things look and work the same online and locally. But the development environment also needs a lot of stuff that the doc environment does not need (we dont need mkl on readthedocs to build the documentation). This change includes the doc/requirements.txt file into the environment.yml to make sure they always use the same packages.

The documenation should be unchanged. Let's see what the conda tester says.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -13,8 +13,5 @@ sphinx:
 formats:
   - pdf
 
-# Optionally set the version of Python and requirements required to build your docs
-python:
-  version: 3.7
-  install:
-    - requirements: doc/requirements.txt
+conda:
+  environment: environment_doc.yml
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,8 +1,0 @@
-sphinxcontrib-bibtex
-recommonmark
-ipykernel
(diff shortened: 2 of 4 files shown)


gassmoeller:
This does not work right now, and I will postpone it until the release is done.

gassmoeller:
Including a pip requirements file used different package versions and broke the tester. So instead I converted the doc/requirements.txt into a conda environment `environment_doc.yml`, and modified the readthedocs workflow to use that conda environment instead of the old `requirements.txt`. The documentation is already build on my readthedocs site, and it looks good as far as I can tell: https://gassmoeller-rayleigh.readthedocs.io/en/unify_doc_and_devel_environment/index.html#.

tukss (approved):
I like the idea of also using conda for the documentation build. The newly built documentation looks good to me. Even the bibliography still works. I'll let you decide when to merge because you said something about waiting until the release.

feathern:
Hey Rene,
Looks good to me too unless I'm missing something.  Is this safe to merge now?
-Nick

PR #295, opened 2021-05-28
Fixing some uninitialized variables found by the intel compiler.

cianwilson:
Compiling using intel with '-check all' stops at runtime on uninitialized variables. This pr fixes some found running a Boussinesq hydro benchmark.  There are likely more.

It would be good if someone else could check my initialization of some of these.  In particular `ref_remember`.


tukss:
I can confirm it fixes the uninitialized variable errors with ifort 19.0.2.187 running the `j2011_hydro_minimal` input.

The initialization of `ref_remember` to `1` looks safe to me, because it won't trigger any other behavior as long as it's not 4 and 1 is the default.

PR #296, opened 2021-05-28
Conda environment file for mac

ksoderlu:
Mac, conda environment: the code would compile, but got the following error when trying to run it:
mpiexec -np 8 ./rayleigh.opt -nprow 4 -npcol 2
Fortran runtime error: Integer overflow when calculating the amount of memory to allocate

This is intended to fix this bug.

PR #297, opened 2021-05-28
Update online documentation for conda environment

gassmoeller:
Simplify the online documentation for the development environment using the environment files. Also add a basemap package needed for the some postprocessing examples (https://rayleigh-documentation.readthedocs.io/en/latest/post_processing/Diagnostic_Plotting.html#II.-Configuring-Your-Python-Environment).

### DIFF
--- a/doc/source/User_Guide/devel_environment.rst
+++ b/doc/source/User_Guide/devel_environment.rst
@@ -11,37 +11,14 @@ Conda Environment
 
 First, if you don't have Conda, you should download and install the version appropriate for your architecture `here. <https://docs.conda.io/en/latest/miniconda.html>`_
 
-Once you have Conda installed, create a Conda environment named (say) radev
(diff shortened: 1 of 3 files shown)


tukss:
The new instructions work for me on a Linux system. Someone else should test the macOS instructions.

Is there are reason why we give the macOS environment the name `radev_mac`? I don't think you can install it at the same time as the Linux environment anyway. The different names might lead to confusion.

feathern:
Could we remove the basemap dependency?    We're going to move to a different shell-slice plotting library that I wrote and which we can manipulate/modify as needed.   It's part of Lydia Korre's shell-slice commit.

gassmoeller:
@tukss: Good catch, that was a leftover from @ksoderlu and my testing in #296. I unified the environment names. I also removed the basemap dependency, the reason we added it was that: https://rayleigh-documentation.readthedocs.io/en/latest/post_processing/Diagnostic_Plotting.html#II.-Configuring-Your-Python-Environment says we need it at the moment.

tukss (approved):
Looks good to me now. I think removing basemap is OK, as we have instructions how to install it later at the right point in the docs.

PR #299, opened 2021-05-28
Python routines to make the lookup table easier to use (and new documentation on it)

orvedahl:
Interfacing with the output lookup table requires knowing the quantity code. This PR allows you to generate a map between quantity codes and quantity names. There is support for including custom diagnostics as well as defining your own names. I also updated the Rayleigh documentation to show how to use the new routines.


orvedahl:
@gassmoeller one of these routines automatically generates a python file, should that be included in the .gitignore so it is not tracked?

cianwilson:
Hi @orvedahl, this looks great.  Particularly like how you've thought about shortcuts and the custom diagnostics.

Couple of questions...
1. The .gitignore now ignores `post_processing/lut_mapping.py` but that isn't necessarily where the `lut_mapping.py` ends up.  When I tried it it seemed to just be written to wherever my current directory was.  Perhaps it should land in `post_processing` when generated with `--default-location`?
2. Is the idea that users should edit `lut_shortcuts.py` with their desired shortcuts?  I think that will also end up creating git trouble as it is tracked.  Perhaps that shouldn't be included?
3. It all works beautifully for me with `--default-location` but when I try to specify a path to the build directory I get:
```
(radev_test) cwilson@fodhuchtu:rayleigh.master$ python3 post_processing/generate_mapping.py src/build
Traceback (most recent call last):
  File "/home/cwilson/repos/github/geodynamics/rayleigh.master/post_processing/generate_mapping.py", line 460, in <module>
    outputQ = OutputQuantities(search_path, default_location=default_loc)
  File "/home/cwilson/repos/github/geodynamics/rayleigh.master/post_processing/generate_mapping.py", line 228, in __init__
    self._parse_diagnostic_files()
  File "/home/cwilson/repos/github/geodynamics/rayleigh.master/post_processing/generate_mapping.py", line 361, in _parse_diagnostic_files
    for Line in mf:
  File "/home/cwilson/conda/miniconda/3.8/envs/radev_test/lib/python3.9/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode bytes in position 40-41: invalid continuation byte
```
That happens both in the conda env and in my native environment.
4. Speaking of the conda env, would it be possible to add `docopt` as a dependency to the `environment*.yml` files?

Not really issues but more discussion points:
1. Perhaps we need to figure out the convention for scripts vs modules in `post_processing`.  Your scripts are added with a `.py` suffix but aren't executable.  There are a couple of scripts in there that also have `.py` but are executable.  And of course there are modules that don't have `__main__` functionality but still end in `.py` and aren't executable.
2. Would it be possible to add a default `lut_mapping.py` to git so that users don't have to generate this themselves.  A post-commit action could then test if running the generation changes anything (if people have added any extra default diagnostics).  I guess this would involve splitting out the custom diagnostics into a separate file/dictionary that your convenience `lut.py` wrapper could merge together.

cianwilson (changes requested):
See my comment:
https://github.com/geodynamics/Rayleigh/pull/299#issuecomment-850666886
which was meant to be submitted as a review but I messed up.

orvedahl:
Thanks for the great feedback @cianwilson 
1) lut_mapping.py should always be placed in the same location as the generate_mapping.py file, because it is imported by lut.py and lut.py assumes it is in the same directory --- this is now the case
2) users are meant to actively change/update lut_shortcuts.py to their hearts content --- so lut_shortcuts.py is now in gitignore
3) I saw the same error, the code did not filter out any of the *.o or *.mod files --- now only Fortran files are parsed
4) I forgot about the docopt package --- this has been added to both environment*.yml files, please make sure I didn't break the MacOS one...

discussion points:
1) I personally moved away from executable python scripts and just do "python my_scipt.py", provided their is an appropriate __main__ section. I would be happy to change my scripts accordingly, but that might be best done all at once in its own PR
2) I like the idea of having a default mapping in the future, especially for the ability to see if quantity codes have changed. I will think about how best to add it, but it might have to wait for its own PR.

orvedahl:
I updated this repo with the latest changes in order to avoid merge conflicts (I modified files that have since been removed)

orvedahl:
I changed how this works from the user's point of view.

1) There is now a default mapping already included (lut_mapping.py), so the user does not need to generate it themselves.
2) The default mapping does not have any system dependent stuff (git hash, url, etc.) in the header, so in theory this can be used as a check to see if quantity codes have changed.
3) The custom root capability is still there: now the user generates the map themselves, which is stored in lut_mapping_custom.py. The lut_mapping_custom.py contains the exact same information as the lut_mapping.py file, but also includes the custom diagnostics and a more helpful header.
4) The user interface (lut.py) will always try to import the custom map first, if that fails it imports the existing default map.

illorenzo7:
@cianwilson @orvedahl Just a reminder to approve and/or merge if you both are happy with this pull request!

Loren

cianwilson (approved):

PR #300, opened 2021-05-28
Activate compiler warnings and remove unused variables

gassmoeller:
This PR activates gfortran warnings ('-Wall') during compilation in debug mode for the automatic tester. There are dozens of active warnings and this PR fixes all that I could easily identify and fix, mostly '-Wunused-parameter' and I think this can be safely merged if the tester passes. 

However, someone in a follow-up PR should go through the remaining warnings and see which ones can be addressed: https://github.com/gassmoeller/Rayleigh/runs/2697219115?check_suite_focus=true

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -23,7 +23,8 @@ jobs:
     - name: Build Rayleigh
       # First make sure notebooks do not contain output, then build doc
       run: |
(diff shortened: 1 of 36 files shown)


cianwilson (approved):
Wow, this must have been tedious to do.  

I've confirmed it builds with these changes on the intel version I have access to.

Looks good!

PR #301, opened 2021-05-28
Fix the versions in the conda environment to ones that are currently known to work

cianwilson:
Only mac so far, will give linux a quick whirl now.

Removed duplicate dependency on sphinx, added in vtk and funcsigs to allow generic input tests to run.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -2,19 +2,21 @@ name: radev
 channels:
   - conda-forge
 dependencies:
-  - make
-  - git
-  - matplotlib
-  - jupyter
-  - scipy
-  - sphinx
-  - sphinxcontrib-bibtex
-  - nbsphinx
-  - pandoc
-  - recommonmark
-  - sphinx
-  - gcc_linux-64
-  - gfortran_linux-64
-  - mkl
-  - fftw
-  - mpich
+  - python=3.9
+  - make=4.3
+  - git=2.30
+  - matplotlib=3.4
+  - jupyter=1.0
+  - scipy=1.6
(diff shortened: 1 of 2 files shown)


environment.yml
```diff
@@ -2,19 +2,21 @@ name: radev
...
-  - fftw
-  - mpich
+  - python=3.9
+  - make=4.3
+  - git=2.30
+  - matplotlib=3.4
+  - jupyter=1.0
+  - scipy=1.6
+  - sphinx=4.0
+  - sphinxcontrib-bibtex=2.2
```
gassmoeller, line 12: Did we resolve the issue with sphinxcontrib-bibtex version larger than 2.0 discussed earlier this week: https://rayleighworkspace.slack.com/archives/C022V39JPMG/p1621870838001100?

If not, should we fix the version to 1.0.0 like @feathern suggested in the slack thread? I somehow lost track of the issue and do not remember our solution.

gassmoeller (commented):
Looks good, thanks for doing this! I have one question, otherwise this looks ready.

environment.yml
cianwilson: I believe we were able to upgrade the version following #269 and no longer pin that package.

I've tested `make doc` on both mac and linux with these files and didn't spot anything going wrong (though I only looked at the pdf).

gassmoeller: Ah thanks, I must have missed #269 somehow. Ready to go then.

gassmoeller (approved):

PR #304, opened 2021-05-28
Build mac env

gassmoeller:
This adds an automatic MacOS tester that builds the Rayleigh conda environment for Mac.

It seems the tester fails, because Mac's default `ld` does not recognize the `--no-as-needed` option. Will investigate further later.


gassmoeller:
I just forgot the Mac configure flag and added that now to the automatic workflow.
I have an idea for how to fix that permanently in another PR, but for now this is good to go.

feathern (approved):
Thanks Rene!

PR #305, opened 2021-05-30
Unify conda environments and configurations on linux and mac

gassmoeller:
This PR sits on top of #304 (so youĺl see it's changes as well in this one) and unifies the environment files and configuration options for linux and mac.

For the environment files there is a meta-package in conda called `compilers` that includes the compiler versions used to build all conda packages indepedent of operating system (so on mac conda will install clang, on linux gcc). I pinned the package version to the most current one, which is also the same compiler version that was previously explicitly selected in the individual packages.

For the configuration script instead of asking every conda Mac-user to add an configuration option, I added a dependence on the $OSTYPE environment variable that is set in nearly all shells (not all, see here: https://stackoverflow.com/questions/394230/how-to-detect-the-os-from-a-bash-script?noredirect=1&lq=1). The exact value in this variable differs a bit between bash/zsh and so on, but always contains `darwin` on MacOS and `linux` on Linux systems. If a user uses an exotic shell that does not set the variable no harm done, the condition will simply not trigger. One thing that  is now no longer possible is to use `no-as-needed` on a Mac system when using conda, but I do not know if that is something that would ever be intended. `mac-mkl` is now somewhat superfluous as option, should we remove it or keep it for compatibility?


feathern (approved):
This looks good to me.   Thanks Rene.  Since the mac-mkl flag is no longer needed, and the compilation looks the same on Linux and Mac, I think we can remove it.  Do you want to do that as part of this PR, or should I go ahead and merge?

gassmoeller:
I can do this as part of this PR, give me a few hours.

feathern:
OK, I'll keep an eye out

gassmoeller:
Ok, done. Looks like the tester likes it as well.

PR #307, opened 2021-06-14
do not install sphinx via apt

tukss:
This breaks the documentation build because the pip version conflicts with the
one from apt. The Ubuntu focal container ends up having an incompatible version
of pygments. The current pip install command pulls in sphinx as a dependency so
there is no need for installing it also via apt.

I am not sure if this fixes the documentation build issues producing all italics because I couldn't reproduce the issue on my system.


gassmoeller (approved):
Yes this makes sense.

PR #308, opened 2021-07-13
Read only some parts of file

illorenzo7:
This changes the Shell_Spectra class to have the ability to read one radial level, one record, and/or one quantity, without reading the whole binary file (uses the offset keyword arg in numpy.fromfile). Interface is unchanged except Shell_Spectra now takes the keyword args (default None)

iitervals
qvals
irvals

For example, Shell_Spectra(fname, path, iitervals=1, qvals=801, irvals=[2,4]).

returns the spectrum of the 2nd record, for the radial magnetic field only, at the 3rd and 5th radial levels. Dimensionality of the vals/lpower arrays will be unchanged. In this case shape (vals) = (nell, nm, 2, 1, 1)

If no new keyword args are specified, behavior is the same as before. 

@tukss and @feathern, I know the new memory-mapping framework may make this obsolete, but in the meantime I think this is a quick way to start reading sections of files quickly. 

I have run some tests and it seems to work as intended, with reading shorter parts of file going much faster than reading the whole file. If other people want to try it out too (and it works for them as well!) it should be straightforward to extend this to the other classes.

### DIFF
--- a/post_processing/rayleigh_diagnostics.py
+++ b/post_processing/rayleigh_diagnostics.py
@@ -1320,6 +1320,8 @@ class Shell_Spectra:
         self.time = self.time[iitervals]
         self.iters = self.iters[iitervals]
         self.qv = qv[iqvals]
+        self.radius = radius[irvals]
+        self.inds = irvals
 
         self.niter = len(iitervals)
         self.nq = len(iqvals)
--- a/post_processing/rayleigh_diagnostics.py
+++ b/post_processing/rayleigh_diagnostics.py
@@ -968,13 +968,22 @@ class Shell_Slices:
             print('.......................')
             print('costheta : ', self.costheta)
 
-    def __init__(self,filename='none',path='Shell_Slices/',slice_spec = [], rec0 = False):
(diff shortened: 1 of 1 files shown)


feathern:
Hey Loren,
Looking at the failed test, it looks like the 'offset' keyword wasn't accepted.   This looks to be a Numpy version issue, since the keyword is new in 1.17.0.    We may want to use fseek instead for backwards compatibility.    Or, better, we could wrap it in a try/except statement:
`try:`  
`    numpy.fromfile(...,offset = X)`
`except:`
`   fd.seek(offset)`
`   numpy.fromfile(... [ with no offset])`

illorenzo7:
Just chatted with Nick about this; it's not quite ready to go, will fix some underlying issues in the coming days, so don't merge yet.

feathern:
OK, for completeness, Loren is referring to the fact that I have a new Shell_Spectra class that might be better to work off of since it has some time-averaging capability for backing up model data.  I'm going to close this pull request now, and you can resubmit later as convenient, Loren.

PR #309, opened 2021-07-21
Python library to perform spectral transforms and derivatives

orvedahl:
This PR provides python routines to compute spectral transforms and derivatives using the same techniques that Rayleigh uses. There are four main classes, each briefly described below.

1) Fourier: provides forward and inverse FFT and a way to calculate phi derivatives from either physical or spectral space.

2) Legendre: provides forward and inverse Legendre transform and a way to calculate theta derivatives from either physical or spectral space.

3) Chebyshev: full support for the radial grids that Rayleigh uses including nonuniform, multiple Chebyshev domains. Radial derivatives and forward/inverse Chebyshev transforms.

4) SHT: full spherical harmonic transform from Shell_Slices to/from Shell_Spectra. Two convenience routines to compute the phi derivative from spectral space and the quantity sin(th)*dF/dth in spectral space.

I have extensively tested these routines throughout their development: 1) run a Rayleigh simulation and output fields and their derivatives as Shell_Avgs, AZ_Avgs, Meridional_Slices, Shell_Slices, and Shell_Spectra. 2) read the data and build the various classes from above. 3) use the python classes to compute the forward/inverse transforms and compare to the Shell_Slices/Shell_Spectra. 4) use the python routines to compute derivatives and compare to the appropriate output.

The python code should be well documented, but I still need to add something to the Rayleigh docs.


orvedahl:
In the process of writing up a jupyter notebook to illustrate the use of these tools, I uncovered a few substantial errors.

Until I track them down, this PR should not be merged.

orvedahl:
I found and fixed the bugs...this should be good to go now.

The jupyter notebook is complete and provides some simple examples of how to use the main tools.

feathern:
Hi Ryan,
This look pretty good to me.  It will be some time before I can test the library out thoroughly, but it looks complete and it doesn't interfere with any other infrastructure, so I'll merge.  Before I do, I have a rewording suggestion.  Can you change the accuracy section paragraph to read:

All classes within the spectral_utils.py module rely on expanding the input data into a series of basis functions.  There are some functions that do not behave well when written as a truncated expansion. Typically, this class of function requires power in all modes of the expansion, up to and including the maximum degree in the truncation. This fact usually leads to underresolved features and a poor final result.

-Nick

orvedahl:
Thanks Nick, the accuracy stuff has been reworded.

feathern (approved):
OK great, this looks good to me then.  Merging...

PR #310, opened 2021-08-12
Allow changing the executable name at the "make install" step

orvedahl:
The default behavior when you do "make target=opt" is to build only the optimized executable and it will be named "rayleigh.opt". Similarly, "make target=dbg" will build the debug version "rayleigh.dbg". The executables are then placed into the bin directory with "make install", but the filenames are always rayleigh.opt or rayleigh.dbg.

When actively developing multiple branches, it can be a little hard to keep track of what executable came from what branch. This PR allows you to change the name of the executable when doing "make install", but only if you requested a particular target.

To build the optimized version and call the executable "branch1.opt", the following would be used:
`make target=opt`
`make target=opt output=branch1.opt install
The resulting executable will be placed into the PREFIX directory, just as before. The added "output=" argument is only used if there is a target specified.

I would view this more of a tool for developers who compile multiple optimized/debug versions at once. The user might find this useful as well, but it grew out of an annoyance while developing multiple branches.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,9 +72,21 @@ clear_ipynb:
 
 .PHONY: install
 install:
+ifeq ($(target), "all")
 	@echo "Installing executables into: " $(PREFIX)"/bin"
 	@mkdir -p $(PREFIX)/bin
 	@cp $(BUILD)/compiled/rayleigh.* $(PREFIX)/bin/.
+else
+ifdef output
+	@echo "Installing rayleigh.$(target) into: " $(PREFIX)"/bin/$(output)"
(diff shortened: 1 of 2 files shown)


feathern:
Hey Ryan,
This looks good to me, but can you also add a small blurb to the build portion of documentation describing this?   I think this is useful.  Thanks. --Nick

orvedahl:
Thanks Nick, I always forget about the sphinx documentation...

feathern (approved):
Awesome.   I think this is very useful.  Approving/Merging...

PR #312, opened 2021-08-23
fix attribute access in reference_tools

tukss:
This interface never worked before. With the fix you can do things like `r.diff_fact = 1.0` and it will call the appropriate setter for constants. It already worked for functions.git push --set-upstream origin fix-set-constant

### DIFF
--- a/post_processing/reference_tools.py
+++ b/post_processing/reference_tools.py
@@ -39,7 +39,7 @@ class equation_coefficients:
         if name in self.f_dict:
             self.set_function(value, name)
         elif name in self.c_dict:
-            self.set_constant[name]
+            self.set_constant(value, name)
         else:
             super().__setattr__(name, value)
 


tukss:
I had to look this up, too. I think we are handling it correctly.
The `f_dict` and `c_dict` dictionaries in the script are giving the Fortran 1-based indices. These are then converted to Python's 0-based indices when using the `set_constant` and `set_function` methods. These can be used with the name or 1-based index.

Example:
```python
rho = np.ones_like(c.radius)
# The next three statements are equivalent.
c.density = rho
c.set_function(rho, "density")
c.set_function(rho, 1)
```

There is no equivalent `get_function` method yet, so you have to do the index conversion manually or just use `c.density` to read the value.

feathern (approved):
Looks good

PR #314, opened 2021-08-27
removed unnecessary Jones 2011 hydro benchmark

illorenzo7:
It's identical to j2011_steady_hydro_minimal, and there is only one hydro case (the steady one) in Jones (2011).


feathern (approved):
Thanks  -- good catch on the duplicate file.  merging.

PR #315, opened 2021-08-27
Remove dimensional flag

illorenzo7:
This flag is no longer used by Rayleigh and will cause the code to crash if not deleted from the "_input" benchmark main_input files.


feathern:
Weird -- I thought this was fixed a long time ago.  Thanks Loren.

feathern (approved):
Merging.

PR #316, opened 2021-08-27
Remove diagnostics energy flux

illorenzo7:
This file is no longer used, since the corresponding quantities are all computed in separate routines.


feathern:
Hey Loren,
I think maybe you submitted this from a branch that wasn't up-to-date?  Why are 71 files changed?  I agree this file should be removed.
-Nick

illorenzo7:
@feathern 

Huh, I understand git better now! I think this should be good to go now. This was an old fix, with a remote branch on my local fork. I had been rebasing off master periodically, but then I'd push and pull from the remote to get things even, which effectively undid my rebase. This time I pushed with -f to the remote, and it should only be 2 commits ahead of master.

feathern (approved):
OK, that's better!  Merging.

PR #317, opened 2021-08-31
add support for LLVM-based AOCC compiler by AMD

tukss:
The options are very similar to gfortran. Some some checks (array
boundaries, FPE etc.) are not supported and have been removed for now.
Signal handling is also not supported at the moment.

This has been tested with AOCC version 3.0.0

### DIFF
--- a/configure
+++ b/configure
@@ -25,6 +25,10 @@ function compiler_version {
   then
     VTMP="GNU"
   fi
+  if [[ $VSTR == *"AOCC"* ]]
+  then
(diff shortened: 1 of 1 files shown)


tukss:
The compiler can be downloaded [here](https://developer.amd.com/amd-aocc/#download) for testing.

feathern (approved):
This looks good to me.   I don't have an AMD compiler, but I did compile via GNU (which shares a lot of logic in your modification) with no problems.

PR #318, opened 2021-08-31
add new configure option -ifort-march

tukss:
This makes ifort use the `-march` option instead of the equivalent `-x`
option. Executables built with the latter option do not run on non-Intel
CPUs. I am keeping this as a nondefault option as I am not sure if there
is a decrease in performance using `-march`.

There is no direct `-march` equivalent for the AVX512 option. Leaving it
as is, as there is no non-Intel AVX512 processor available at the time
of writing.

### DIFF
--- a/configure
+++ b/configure
@@ -211,7 +211,11 @@ function displayhelp {
     echo "      library.  Set this flag to link statically instead."
     echo ""
 
-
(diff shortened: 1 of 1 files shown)


feathern (approved):
This looks good to me.  Thanks Philipp.

PR #319, opened 2021-09-07
documentation for multiple radial domains

### LINKED ISSUE #4: Multiple Chebyshev Domains (Documentation)
This functionality needs to be documented.

tukss:
This is a simple documentation of the multiple radial domain feature. I don't know any of the internals, this is just from what I figured out by reading the namelist definitions. Review by someone more knowledgeable is appreciated.

I also changed the way the namelist variables are formatted in this section. In my opinion math mode is not a good choice for variable names, in particular because underscores are not formatted correctly. I'm using code formatting (enclosed in ` `` ` marks) instead.

Closes #4.

### DIFF
--- a/doc/source/User_Guide/running.rst
+++ b/doc/source/User_Guide/running.rst
@@ -206,6 +206,28 @@ Note that the interpretation of :math:`rmin` and :math:`rmax` depends on
 whether your simulation is dimensional or nondimensional. We discuss
 these alternative formulations in §\ :ref:`physics`
 
+It is possible to run Rayleigh with multiple, stacked domains in the
+radial direction. Each of these is discretized using their own set of
+Chebyshev polynomials. The boundaries and number of polynomials can be
+set for each domain indiviadually, which makes it possible to control
+the radial resolution at different radii.
+
+To use this feature the problem size has to be specified using
+``domain_bounds`` and ``ncheby`` instead of ``rmin``, ``rmax``, and
+``n_r``. ``ncheby`` takes a comma-separated list of the number of
+polynomials to use in each domain. ``domain_bounds`` takes
+a comma-separated list of the radii of the domain boundaries. It has one
+element more than the number of domains. This is an example of two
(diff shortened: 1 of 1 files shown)


illorenzo7:
Hi @tukss , this looks mostly good! Just a few things (@feathern maybe you should make sure I have this right)

1. ncheby is actually the number of radial points (not polynomials) in each domain (i.e., the variable name is a bit of a misnomer). I believe the number of polynomials is dealised in the same way as n_theta --> l_max, i.e., floor (n_r/1.5) so in your example there would be 10, 42 chebyshev polynomials. 
2. Possibly emphasize the fact that although the radial array in the diagnostics/code is DECREASING, the domain_bounds are listed in increasing order
3. Possibly mention that the radial array in the diagnostics/code will contain repeated radii for internal domain boundaries, for which most quantities are forced to be continuous.

feathern:
Ooh, thanks @tukss .  I will try hard to have a careful look at this tomorrow (Friday). -Nick

tukss:
Thanks for pointing these things out @illorenzo7! I changed the text accordingly.

And thanks @feathern for looking into this tomorrow.

illorenzo7:
@tukss looks good! It still says "polynomials" instead of "points" before the example (which is technically true, I guess, before dealiasing), but anyway, happy to merge after @feathern takes a look. Thanks for this!

Loren

illorenzo7 (approved):

tukss:
> It still says "polynomials" instead of "points" before the example (which is technically true, I guess, before dealiasing), but anyway, happy to merge after @feathern takes a look.

Thanks. I changed that, too.

feathern:
OK, I'm good with this for now.  I will try to get an update in during the next few days.  This documentation is correct, but also incomplete.  As the one who wrote the multi-cheby code, though, I should make it complete :)

Also, we probably don't want to say 'most variables are continuous...'  :)  I'll fix that too!

PR #320, opened 2021-09-08
Small bug (missing pathname) for equation_coefficients file

illorenzo7:
The equation_coefficients file was going to the parent directory in ensamble mode, due to a missing "my_path" in front of the file name.

### DIFF
--- a/src/Physics/PDE_Coefficients.F90
+++ b/src/Physics/PDE_Coefficients.F90
@@ -798,7 +798,7 @@ Contains
         If (present(filename)) Then
             ref_file = Trim(my_path)//filename
         Else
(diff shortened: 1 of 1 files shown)


tukss:
This looks correct to me. I am wondering if the same fix should be applied to the subroutine `Read_Custom_Reference_File` (line 801). Otherwise it won't read the `reference` file from the right directory if a specific filename is not provided.

illorenzo7:
@tukss

Thanks for noticing this! I went ahead and fixed the line you pointed out as well. I think this pull request is good to go now.

feathern (approved):
Awesome.  Thanks for getting this one finished up!

PR #323, opened 2021-09-24
Remove unnecessary Jenkinsfile

gassmoeller:
We deactivated Jenkins and use Github Actions instead. Now the Jenkinsfile no longer needed.


feathern (approved):
Looks OK to me.  Always good to keep things tidy!

PR #327, opened 2021-10-20
Ohmic Heating Prefactor Fix (output bug)

feathern:
Loren pointed out that the ohmic heating terms in Rayleigh's output (not in the main code loop) were missing a factor of density*temperature similarly to the viscous heating term.   This PR adds that factor.
-Nick


illorenzo7 (approved):
This one looks good too! Thanks :)

PR #329, opened 2021-10-21
Point_Probes docstring update

feathern:
I just realized that the rad_inds and theta_inds attributes of the Point_Probes class were not documented in the doc_string.   The phi_inds attribute was documented, but mislabeld as phi indices.  This  PR fixes these three errors.


illorenzo7:
@feathern OK sounds good! I'll hold off and also do the Meridional_Slices/Shell_Slices stuff after this pull request.

feathern:
Merging now to get this out of the way.  Thanks Loren.

PR #334, opened 2021-10-29
Removed unused lines in read_custom_reference_file

feathern:
Read_custom_reference_file was originally written as though the input parameter 'filename' was optional.   In practice, that filename is always supplied, making a few lines of code in this routine obsolete.  This PR removes those lines of code.


orvedahl (approved):
This looks good, I'll merge it once the checks are complete.

PR #335, opened 2021-10-29
Advect ref

feathern:
This commit sets advect_reference_state to true by default.  As discussed in the issues section, this is probably the more intuitive default setting.  Note that with this commit, Rayleigh's internal benchmark routines now explicitly set this variable to false if one of the four benchmark modes are activated.   All input files have this variable set to false for consistency with the old default value that they assumed.

### DIFF
--- a/input_examples/benchmark_diagnostics_input
+++ b/input_examples/benchmark_diagnostics_input
@@ -16,6 +16,7 @@
  magnetism = .false.
  viscous_heating = .false.
  ohmic_heating = .false.
+ advect_reference_state = .false.
 /
 &temporal_controls_namelist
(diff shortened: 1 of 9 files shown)


illorenzo7 (changes requested):
Thanks Nick! I think this looks basically good, but did you forget to change advect_reference_state to .true. in Restore_Physics_Defaults() (line 213 of Controls.F90)?

feathern:
Good catch.  I just updated the PR.

illorenzo7 (approved):

illorenzo7:
Great! All looks good now. Merging.

PR #337, opened 2021-11-01
Shell slice and spectra consistency

illorenzo7:
Makes doc string (regarding indices) compatible with new doc strings for Point_Probes and Meridional_Slices. Also gives Shell_Slices, Shell_Spectra, SPH_Modes, and Power_Spectrum the alternate property

rad_inds (for compatability with Point_Probes)

identical to 

inds (for backwards compatability). 

There used to only be the property inds.

### DIFF
--- a/post_processing/rayleigh_diagnostics.py
+++ b/post_processing/rayleigh_diagnostics.py
@@ -730,7 +730,7 @@ class Point_Probes:
                                                   : corresponding to each point in self.radius
     self.theta_inds[0:self.ntheta-1]              : theta indices (from the full simulation theta grid) 
                                                   : corresponding to each point in self.costheta
-    self.phi_inds[0:self.nphi-1]                  : phi indices (from the full simulation theta grid) 
+    self.phi_inds[0:self.nphi-1]                  : phi indices (from the full simulation phi grid) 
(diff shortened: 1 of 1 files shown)


illorenzo7:
Note: I tested this for Power_Spectrum, Shell_Spectra, and Shell_Slices, but I don't have any sample SPH_Modes (and I am also unfamiliar with that output). If someone wants to test on SPH_Modes that would be great!

feathern:
Hi Loren,
This looks good to me.  Just to be sure, though, can you confirm that you read in sample files for those classes whose init routines you modified (not worried about the docstrings).
-Nick

illorenzo7:
I haven't checked the SPH_Modes and it looks like there are multiple ways to initialize the Power_Spectrum. I will check on this now and leave another comment, hopefully shortly.

feathern:
OK thanks.  I suspect it's fine, but just to be safe it's good to run through a few read-ins.

illorenzo7:
OK, I just checked the SPH_Modes and other (blank and power_file) options to initialize Power_Spectrum. It seems good on my end!

feathern (approved):
Great.  Approved.

PR #338, opened 2021-11-05
fix error in CPU autodetect message

### LINKED ISSUE #302: selecting native on mac using the conda environment gives a non-fatal error
configure uses:
`gfortran -march=native -Q --help=target`
for gnu builds when native is selected at configure time.

Using the conda mac environment this results in:
```
fortran: fatal error: cannot execute 'cc1': execvp: No such file or directory
compilation terminated.
```
which looks scary but the configuration actually carries on.

This appears to be caused by cc1 not being packaged by conda.  Anyone know of a workaround?

tukss:
This should fix an error message displayed if the GNU compiler is used,
but `gfortran` is not a valid compiler. I don't see a downside to this
change because we already make sure `FVERSION` is set to `GNU`.

Fixes #302.

### DIFF
--- a/configure
+++ b/configure
@@ -752,7 +752,7 @@ then
        fi
        FFLAGS[0]="${OPT_FLAGS} -march=${opt}"
        if [[ "$opt" == "native" && $FVERSION == "GNU" ]]; then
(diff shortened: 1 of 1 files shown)


feathern (approved):
Great.  Approved!

PR #345, opened 2021-11-19
fix quotation marks in namelist

tukss:
The docs had a wrong quotation mark in the Fortran namelist.


illorenzo7 (approved):
Looks good, thanks!

PR #346, opened 2021-12-03
[Critical]  Fix CN implementation of radially varying magnetic diffusion

feathern:
The benchmarks, as well as most users, run with diffusion coefficients (nu, kappa, eta) that are constant in radius.   Because of this, a bug in the code has gone unnoticed for a very long time.   The diffusive terms are evolved using the semi-implicit Crank-Nicholson method, which is effectively an average of the forward and backward Euler methods.   A magnetic diffusion term for the A-potential equation, namely dA/dr*deta/dr was being evolved without the explicit contribution from the CN scheme (i.e., the forward Euler component).   This PR fixes the issue.

Please do not accept this PR just yet.  I would like to hear from Rene about how we might automatically manage release notes before this is fully accepted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - None yet
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


feathern:
This PR is safe to accept now.  I also added the changelog file as discussed.

orvedahl (approved):
This looks good to me, thanks Nick.

PR #349, opened 2021-12-13
Automatic calculation of dTdr_top for models with internal heating.

feathern:
This PR modifies the code so that the user no longer needs to specify dTdr_top when running models with internal heating and fix_dTdr_top set to true.   This process is very error-prone from a user standpoint, and so it is now done automatically.  Any additional flux passing through the lower boundary is also taken into account in the event that fix_dTdr_bottom is set.   The behavior may be overridden by setting adjust_dTdr to false in the Boundary_Conditions namelist.


illorenzo7 (approved):
Nice! This looks good, and very useful. It took me a few minutes to follow the signage, but it looks correct. Merging.

PR #351, opened 2021-12-29
update VERSION file

tukss:
The `VERSION` file had not been updated to reflect the new release. At the moment this file is only used when generating the PDF version of the documentation.


feathern (approved):
Good catch - merging.

PR #352, opened 2022-01-11
Main input Class

feathern:
This commit adds a main_input class to rayleigh_diagnostics.py.  I have found this useful when designing parameter space sweeps.   A notebook describing how to use the class has been included in this commit.


cianwilson (changes requested):
This looks great @feathern !

Just one possible fix in the code and one comment.

One other quick thought...  Since dictionaries aren't ordered is there a risk here of making main_input files hard to diff if the output order changes.  Would it be worth considering using OrderedDicts instead?  On the namelist level this could also be achieved by keeping the order of the namelists fixed from the time of input but this is currently overridden when a new option is force set.

feathern:
Thanks Cian,
I changed the dicts to OrderedDicts and added that missing write statement (good catch).   I tried changing the loop, but something a little further down crashed, and I'm in a bit of a hurry.  So I'll leave that as it is for now.
-Nick

cianwilson (approved):
Looks great!

PR #354, opened 2022-01-20
add a spack package

tukss:
This adds a spack repository with a package for Rayleigh and the corresponding documentation. It's quite useful for building Rayleigh in HPC environments because it takes care of building all the dependencies with the right compiler, MPI etc. It can also be used to provide Rayleigh as a module to share with other users.

### DIFF
--- a/doc/source/User_Guide/installation.rst
+++ b/doc/source/User_Guide/installation.rst
@@ -85,7 +85,7 @@ described above, if Intel is used and *all* is selected, every version will
 be compiled. To build only a single version, the **target=<target>** option
 may be used at the **make** stage, for example:
 
-#. **make target=opt** - build only the optimzed version, **rayleigh.opt**
+#. **make target=opt** - build only the optimized version, **rayleigh.opt**
 
 #. **make target=dbg** - build only the debug version, **rayleigh.dbg**
 
@@ -105,9 +105,9 @@ option) will show a new file named **a.out**.
 If both the optimized version and the debug version have already been built, they
 can be renamed at install time as:
 
-#. **make** - build both optimized and debug verion (or all versions)
+#. **make** - build both optimized and debug version (or all versions)
 
-#. **make target=opt output=a.out.opt install** - install and rename the optimzed version
+#. **make target=opt output=a.out.opt install** - install and rename the optimized version
 
 #. **make target=dbg output=a.out.dbg install** - install and rename the debug version
 
--- a/doc/source/User_Guide/installation.rst
+++ b/doc/source/User_Guide/installation.rst
@@ -138,7 +138,7 @@ The next step has only to be performed once to add the Rayleigh package reposito
 
 .. code-block:: bash
 
-    spack add spack-repo
+    spack repo add spack-repo
 
 Afterwards you can just install Rayleigh and its dependencies using:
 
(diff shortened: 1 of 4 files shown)


doc/source/User_Guide/installation.rst
```diff
@@ -123,3 +123,45 @@ installation is performing correctly. If you are running Rayleigh for
...
+Spack is a package management tool designed to support multiple versions and
+configurations of software on a wide variety of platforms and environments. It can be used to build Rayleigh with different compilers and a custom set of libraries for MPI, LAPACK, and FFTW. It can automatically build dependencies itself or use those provided by the HPC environment.
+
+To set up Spack in your environment follow the instructions in the `documentation <https://spack.readthedocs.io/en/latest/getting_started.html>`_. Add local `compilers <https://spack.readthedocs.io/en/latest/getting_started.html#compiler-configuration>`_ and `packages <https://spack.readthedocs.io/en/latest/getting_started.html#system-packages>`_ as desired.
+
+The next step has only to be performed once to add the Rayleigh package repository. Run this from the base directory of the Rayleigh repository.
+
+.. code-block:: bash
+
+    spack add spack-repo
```
cianwilson, line 141: Should this be:
```
spack repo add spack-repo
```
?

cianwilson (commented):
This is great @tukss,
I've had no issue getting this going with gcc on a number of machines.  I'm still struggling a bit with intel.  Running:
```
spack install rayleigh %intel@18.0.1
```
(on a machine where I've had spack find that version of the intel compilers) completes the build but fails to run with:
```
error while loading shared libraries: libmkl_intel_lp64.so: cannot open shared object file: No such file or directory
```
May have nothing to do with this pull request but I'll keep investigating.

One quick question, in setting up the development environment is it building whatever version of rayleigh is in the directory you supply it with `-p` and is the name of the spack build (`rayleigh@master` in the docs) arbitrary?

Not a part of this PR but this might be a good impetus to start installing everything specific to a version of rayleigh, like the contents of `pre|post_processing` into a suitable PYTHONPATH.

cianwilson:
> May have nothing to do with this pull request but I'll keep investigating.

Actually, straight after commenting on this I managed to get a build explicitly including intel-mkl, as you suggested in the docs, working:
```
spack install rayleigh %intel@18.0.1 ^intel-mkl
```

doc/source/User_Guide/installation.rst
tukss: Thanks for spotting this. I pushed a fix.

cianwilson (approved):
Thanks for fixing the docs!  Looks great.

Still curious about clarifying, when you set up the development environment as described in the docs, is it building whatever version of rayleigh is in the directory you supply it with -p and is the name of the spack build (rayleigh@master) arbitrary?

tukss:
> Still curious about clarifying, when you set up the development environment as described in the docs, is it building whatever version of rayleigh is in the directory you supply it with -p and is the name of the spack build (rayleigh@master) arbitrary?

It will build whatever is in that directory. The version is supplied for two reasons.
1. If the directory after `-p` doesn't exist yet, that version will be put in that place.
1. The version is used to register the compiled package in the database. This is particularly important if other packages depend on it.

cianwilson:
Thanks @tukss !

PR #355, opened 2022-02-07
Revised build_file_list routine

feathern:
This small PR updates the build_file_list routine as follows:
1) A doc string has been added
2) Non-numbered files found by listdir are now ignored (previously these would cause a crash).
3) Logic related to the '_special' suffix has been removed.  This dates back to an older version of this routine I developed for use with ASH data and is irrelevant to Rayleigh's file format.


lkorre (approved):
This looks good to me!

feathern:
Thanks for the quick look.  I'll merge now.

PR #356, opened 2022-02-18
Updated G_Avgs class

feathern:
This is a sizeable update to the G_Avgs class in rayleigh_diagnostics.py.
Changes are:
1)  G_Avgs now supports concatenation of multiple files (via passing a list to the class).  Previously this had to be done via a user-programmed loop outside of the G_Avgs methods.
2)  G_Avgs files can now be output (useful for writing a single G_Avgs file containing the data from many concatenated files).
3)  After the header is read, the entire G_Avgs file structure is now read in at once, instead of record by record.  On some file systems this can lead to a substantial (10--100x) speedup when reading a large series of G_Avgs files.
4)  Individual quantity codes may now be extracted from the G_Avgs file.

Full documentation has been provided via docstrings for all new methods added.

-Nick

### DIFF
--- a/post_processing/rayleigh_diagnostics.py
+++ b/post_processing/rayleigh_diagnostics.py
@@ -22,6 +22,7 @@ from __future__ import print_function
 import numpy as np
 import os
 import glob
+from collections import OrderedDict
 
 maxq = 4000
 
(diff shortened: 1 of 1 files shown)


feathern:
This has languished for over a month.  Merging.

PR #357, opened 2022-02-18
A fix for spectral input when trying to generate input from purely radial functions

cianwilson:
Errors in the assumed size of l_max and m_max when converting purely radial functions would lead to index errors.  Hopefully some minor changes to the script fixes this.

Most of the rest of this PR just adds tests that compare purely radial input hard-coded into rayleigh (init_type 7) with sparse and dense generic inputs generated with scripts.


feathern (approved):
This looks good to me.  I just tested it with my own sparse profile.  Seems good!
Sorry this took so long to review.

PR #358, opened 2022-03-17
Custom Initial Entropy Profile

feathern:
This commit adds functionality.   Rayleigh already possesses an initial condition option to specify a conductive entropy/temperature profile + random thermal perturbations.  This commit extends that functionality to allow a user-specified ell=0 profile to be used instead.   Documentation and an example notebook are included.    This is accomplished via the generic input interface, so I would suggest Cian review this.


feathern:
Been about 4 weeks and no review.  merging.

PR #360, opened 2022-03-23
Fix formula for computing l_max from n_theta

orvedahl:
The number of spectral modes and the number of physical grid points are related as: n_theta = 3 * n_l / 2, where n_l = l_max + 1. This means: l_max = (2 * n_theta / 3 ) - 1.

Previously, if the user provided n_theta, then Rayleigh would compute l_max = (2*n_theta - 1) / 3, which is incorrect. This PR fixes the formula.


orvedahl:
This should not be merged until I figure out why the CI tests are failing.

orvedahl:
This should be good to go now. Note, I did have to update the benchmark results (numbers changed in the last few digits).

feathern:
Hi Ryan,
Looks good. Before I merge, I want to double check that you ran the full benchmark and that the results look correct?  Having to change a digit or two in the partial benchmark check, used for the CI, makes complete sense given that the lmax has changed by 1.  Just want to verify that a full benchmark has passed too (can't see why it wouldn't).
-Nick

orvedahl:
Yes, I ran a full benchmark and the results were as expected.

PR #363, opened 2022-04-09
Improved version of 2a3a and 3b2b transposes

orvedahl:
This PR provides an option to use slightly faster versions of the 2a3a and 3b2b transposes. The original transpose routines (now the default) assumed that the m_values were stored as high/low pairs as: m = 0, m_max, 1, m_max-1, 2, m_max-2, .... Then various pairs of m values are distributed across rows/columns. For a process that holds the first 3 pairs of m values, the order would be: m=0,m_max,1.m_max-1,2,m_max-2. This means accessing the m values in order is not contiguous leading to a small performance hit.

The new 2a3a transpose routine relies on a different storage of the m values, such that the local chunk of m values is contiguous. For a process that holds the first 3 pairs of m values, the ordering will now be: m=0,1,2,m_max,m_max-1,m_max-2. This allows for contiguous access of the m values.

The new 3b2b transpose routine simply reorders some of the for loops to take advantage of the Fortran layout, i.e., column major loop ordering.


feathern (approved):
Approved -- thanks for this too Ryan.  Good bit of optimization here.

feathern:
There was a small conflict in changlog.md that I resolved.   I'm merging this now.

PR #364, opened 2022-04-29
Strict_L_Conservation logical bug fix

feathern:
This small PR fixes a bug that allowed Strict_L_Conservation to be set to True when no_slip_boundaries or no_slip_top were also set to True.   Physically, this flag is inconsistent with no_slip_boundaries.  Numerically, it overrides the no_slip_top boundary condition, so is also incompatible with that BC.   Note that Strict_L_Conservation IS compatible with no_slip_bottom, though that use case would be rare.

Through this bug fix, no_slip_boundaries and no_slip_top take precedence over Strict_L_Conservation.  A warning message indicating that this is happening is produced if a conflict is detected at runtime.


feathern:
Just a note to @tukss -- this is ready for review now.  I fixed my logic issue.

feathern:
Small comment on this.  I also modified a line related to one of Ryan's PR in the changelog.  That was filed as an addition, but it was really a bug fix.

tukss:
I've just tried it out and the logic works here, showing the expected result in `jobinfo.txt`. The tests should hopefully pass with the #365 merged.

Just a semantic remark: The message we send to the user is called `Error`, but we actually treat it more as a `Warning` and the code continues to run with modified settings. In other parts of the code `Error` actually aborts the run. Should we rename this?

feathern:
I will change to 'Warning.'  That occurred to me as well.   Also, I just pulled an update from master, then repushed my changes.  Hopefully taht will force the new conda environment to be used.  The tests appeared to be failing for the same reason before I did that.  We'll see how this goes.

feathern:
@tukss OK, the checks just passed, and then I changed the wording.   Assuming these checks pass, I think you're good to merge/approve.  Only thing I could imagine is that I goofed a quotation mark.

src/Physics/BoundaryConditions.F90
```diff
@@ -93,14 +94,43 @@ Subroutine Initialize_Boundary_Conditions()
...
 
         If (no_slip_boundaries) Then
             no_slip_top = .true.
             no_slip_bottom = .true.
         Endif
-        stress_free_top = .not. no_slip_top
+        stress_free_top    = .not. no_slip_top
         stress_free_bottom = .not. no_slip_bottom
 
+
```
tukss, line 131: I think these blank lines were added by accident.

tukss:
Other than the blank lines it looks good to me. I'll approve already. Feel free to merge having removed the lines or not.

tukss (approved):

PR #366, opened 2022-08-20
fix use of undeclared variable in OMP PARALLLEL clause

tukss:
The Cray compiler (cce) is giving an error because we are mentioning the undeclared variable `nthr` in the `OMP PARALLEL` clause. It isn't being using if `useomp` is not set in the preprocessor, which is probably why we haven't seen problems with the other compilers. I see several ways to fix this. Which is the best depends on our handling of OpenMP, which I don't know much about.

1. Declare `nthr` and set it in both parts of the `#ifdef useomp`. Actually we can use one consistent definition of `kstart` and `kend` here, which makes the code cleaner in my opinion. _This is what I implemented in this PR so far._
2. Remove `nthr` altogether and call `omp_get_num_threads()` twice to the the number of threads. These calls should be very cheap.
3. Define `nthr` locally within a Fortran `BLOCK`.
4. Move the whole `OMP PARALLEL` clause into the `#ifdef useomp` and capture the `OMP END PARALLEL` accordingly. This might have an impact on compiling the code with the OpenMP option being used but `useomp` being unset. I'm not sure we support this anyway, because the other `OMP` clauses would need to be `#ifdef`ed as well.

I don't have a particular preference. Maybe @feathern can make a call here.

### DIFF
--- a/src/Math_Layer/Chebyshev_Polynomials.F90
+++ b/src/Math_Layer/Chebyshev_Polynomials.F90
@@ -568,7 +568,7 @@ Contains
         Integer, Intent(In)    :: ind, dind, dorder
         Real*8, Allocatable :: dbuffer(:,:,:)
         Integer :: dims(4), n,n2,n3, i,j,k, order
-        Integer :: kstart, kend, trank
+        Integer :: kstart, kend, trank, nthr
         Integer :: nglobal, nsub, hoff, hh
         dims = shape(buffer)
         nglobal = dims(1)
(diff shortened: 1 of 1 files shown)


feathern (approved):
Thanks Philipp,
I wonder if this variable has always been missing.  We used to use OpenMP on Mira quite a bit, but that was a long time ago -- maybe it got lost in a development push and we never caught it due to the preprocessor directives.   Approving.
-Nick

PR #370, opened 2022-09-12
Automated Version Update in configure

nicholasjnelson:
Added code to read the version number used in configure from the VERSION file


gassmoeller (approved):
Makes sense to me, but I would like @tukss to take another look at it.

PR #372, opened 2022-09-12
Use Mollweide Projection for Shell Slices

nicholasjnelson:
Changed plot_Shell_Slices.py to make Mollweide plots

### DIFF
--- a/post_processing/plot_Equatorial_Slices.py
+++ b/post_processing/plot_Equatorial_Slices.py
@@ -97,7 +97,7 @@ sizetuple=(8,5)
 fig, ax = plt.subplots(figsize=(8,8))
 tsize = 20     # title font size
 cbfsize = 10   # colorbar font size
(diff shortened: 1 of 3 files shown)


tukss:
Just as a comment: the new Python interface from #274 will also support Mollweide and other projections from matplotlib natively.

tukss (changes requested):
Can you remove your changes to `YT_volrender_tutorial2.ipynb` from the PR? This is just some changed metadata, that we don't want to change every time someone commits something. I can show you how to remove it from history with `git rebase` if you want.

The Mollweide plot works totally fine. We might want to pass `rasterized=True` to the `pcolormesh` call. Especially for high resolution data, this really reduces the size of the PDF file. 192 kB vs. 12 MB in my case.

tukss:
We need to rebase this fix the build error.

tukss (approved):

PR #374, opened 2022-09-12
Reading with byte swapping should be used to read self.version, self.…

hirokemono:
If the global average data G_Avgs with different endian,  self.version, self, niter, self.nq flooring with "bcheck" should be swapped the byte order.


tukss (approved):
This looks correct to me. I verified it still works fine when reading data using the same endianess.

PR #376, opened 2022-09-12
Fixing version number in configure file

jorafb:
From 0.9.5 to 1.1.0


feathern (approved):
Great - glad it works now.

PR #377, opened 2022-09-12
Automated Version Update in configure

nicholasjnelson:
Added code to read the version number used in configure from the VERSION file

### DIFF
--- a/configure
+++ b/configure
@@ -455,7 +455,12 @@ do
 done
 echo ""
 echo " -----------------------------------------------------------------------------"
-echo "                          Rayleigh v1.1.0 Configuration"
+version_file = "VERSION"
+while read -r line
+do
+  version_str = "$line"
+done < "$version_file"
+echo "                          Rayleigh v"$version_str" Configuration"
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -455,7 +455,12 @@ do
 done
 echo ""
 echo " -----------------------------------------------------------------------------"
-echo "                          Rayleigh v1.1.0 Configuration"
+version_file = "VERSION"
+while read -r line
+do
+  version_str = "$line"
+done < "$version_file"
+echo "                          Rayleigh v"$version_str" Configuration"
```
gassmoeller, line 463: To comply with some coding conventions in bash and to make it a bit simpler I would change this slightly to the following:
```suggestion
VERSION_FILE="VERSION"
VERSION=`cat $VERSION_FILE`
echo "                          Rayleigh v$VERSION Configuration"
```
Can you check if this works too?

feathern:
Nick implemented the change of capital vs. lower-case letters.   That appears to work fine.  The build is failing on creating the documentation, which is unrelated to this PR...

PR #378, opened 2022-09-12
Switch Sphinx theme.

gassmoeller:
Remade version of #373 to trigger the readthedocs build for this PR. I will need to play around with some options, so I will keep this as work in progress for now.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,6 +21,9 @@ jobs:
     - name: Build documentation
       # First make sure notebooks do not contain output, then build doc
       run: |
+        # this is to fix GIT not liking owner of the checkout dir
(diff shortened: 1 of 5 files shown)


gassmoeller:
Ok, it looks like I fixed all the bits and pieces that came up. Additional fixes I needed:
- Fix some versions in the docker container
- Fix a github action issue
- Switch the Dockerfile back to OpenMPI (from MPICH). The container on docker hub actually always used OpenMPI, I do not know why the Dockerfile was set to MPICH. I will work on update the containers anyway later this week so this shouldnt be an issue.

TLDR: Check if this looks good, and if it does this is ready to merge. New documentation layout:
https://rayleigh-documentation--378.org.readthedocs.build/en/378/

feathern (approved):
Looks great!

PR #379, opened 2022-09-13
Rename and clarify input and output sections.

gassmoeller:
A suggestion for a renaming of two of the top level section to make their purpose more clear.


feathern (approved):
I'm good with this.  Definitely clearer.

PR #380, opened 2022-09-13
Move some documentation sections into the user guide

gassmoeller:
This is a suggestion for how to reduce the number of top-level sections in the documentation. I think all three of these are really sub-parts of the user guide. I am not entirely sure where to put them, but we can easily reorder the sections in the user guide. If you want me to put them in a different place I can move them around.


feathern (approved):
This looks like a good start to reorganization.

PR #382, opened 2022-09-13
Out-of-Buffer Chebyshev Derivative Routine

feathern:
The radial derivative routine typically computes and stores derivatives within the same buffer.  
If v_r is stored in input buffer(:,:,:,ind), for instance, then dvrdr would be stored in  input buffer(:,:,:,dind).

This add an out of buffer routine, so that now dvdr is stored in another buffer, obuffer(:,:,:)


cianwilson (approved):
Looks good!

PR #383, opened 2022-09-13
allow plot_G_Avgs.py to parse files with variable number of iterations

### LINKED ISSUE #375: Reading Failure a series of global average data files with different sizes (self.niter)
The Global average data reading class (G_Avgs in Rayleigh_diagnostics.py), the size of data array "gavgs" seems to be defined by using the number of steps in one data file self.niter in the first global average data file. However, the the data series in INCITE data has a different size of self.niter, and program plot_G_Avegs.py crushes. The data array to read would be defined when reading each data file as the old version.

File to be read:  G_Avgs/06560000
self.version, self.niter, self.nq: 3 100 14
File to be read:  G_Avgs/06570000
self.version, self.niter, self.nq: 3 41 14
Traceback (most recent call last):
  File "/Volumes/pub_s/INCITE_summary/ryadav/ensemble2/Pm0.1_Pr1_eta0.35_Ek1E-6_Ra1E10_no-slip_dyn/plot_G_Avgs.py", line 112, in <module>
    gavgs[i0:i1,:] = a.vals
ValueError: could not broadcast input array from shape (41,14) into shape (100,14)

tukss:
This resizes the array on the fly to support G_Avg files with a variable number of iterations per files.

fixes #375

### DIFF
--- a/post_processing/plot_G_Avgs.py
+++ b/post_processing/plot_G_Avgs.py
@@ -99,20 +99,27 @@ a = G_Avgs(filename=files[0],path='')  # Here, files[0]='G_Avgs/00010000'
 
 # Loop over all files and concatenate their data into a single array
 nfiles = len(files)
+i0 = 0
 for i,f in enumerate(files):
     a = G_Avgs(filename=f,path='')
+    nq = a.nq
+    niter = a.niter
     if (i == 0):
-        nq = a.nq
-        niter = a.niter
-        gavgs = numpy.zeros((niter*nfiles,nq),dtype='float64')
-        iters = numpy.zeros(niter*nfiles,dtype='int32')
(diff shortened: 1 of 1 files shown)


hirokemono:
It still fails after reading the send file because arrays seem s to be expanded by using nsize from the first data file.

a.niter would be used to expand array size in lines 112 to 116 instead of nsize.

        gavgs = numpy.resize(gavgs, (gavgs.shape[0] + a.niter, gavgs.shape[1]))
        iters = numpy.resize(iters, iters.shape[0] + a.niter)
        time = numpy.resize(time, time.shape[0] + a.niter)

    i1 = i0 + a.niter

tukss:
The new commit actually updates `niter`. Thanks for finding it @hirokemono.

hirokemono (approved):
It works! thanks!

PR #384, opened 2022-09-13
Pscalar

jorafb:
I think the sign of the buoyancy term associated to the chi variable should be opposite to the sign of the linear buoyancy term associated to temperature (if Chi represents composition)


cianwilson:
Yes, that's a good point.  At the moment chi is designed to do exactly what the thermal variable does so we should fix this.

A couple of things:
* this shouldn't be merged into `master` (so please don't anyone merge this yet) but instead into the `pscalar` branch of the `dtm-geodynamics` fork - this isn't yet suitable for merging into `master` because it will break everything and force everyone to use an extra scalar field.  Could you propose the pr into that fork instead please?
* the sign change has caused the `chi_scalar` test to fail so that needs to be fixed (hopefully just by changing the chi_Rayleigh_Number sign but checking this will also flag if there's anywhere else in the code that needs to be fixed)

Thanks!

cianwilson:
This has now been moved to a PR under dtm-geodynamics...
https://github.com/dtm-geodynamics/Rayleigh/pull/1
Closing...

PR #385, opened 2022-09-14
Add workflow that creates a docker container of Rayleigh

gassmoeller:
This adds a github action workflow to automatically create a docker image of Rayleigh when:
1. A PR is merged into master. A new docker image with the tag geodynamics/rayleigh:latest is generated.
2. A Rayleigh release is generated. A new docker image with the tag geodynamics/rayleigh:RELEASE_TAG is generated (where RELEASE_TAG is replaced by the tag of the release used by github).

The image that is generated itself is a bit outdated at the moment (it is based on the tester container using Ubuntu 18.04), but the workflow will use the Dockerfile in docker/rayleigh to build the image, so updating the Dockerfile will automatically update the docker image.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,54 @@
+name: github-docker
+
+on: 
+  push: 
+    branches:
+      - 'master'
+  release:
+    types: [created]
+
+concurrency:
+  group: docker-build
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  build-docker:
+    runs-on: ubuntu-latest
+    if: github.repository == 'geodynamics/rayleigh'
+    steps:    
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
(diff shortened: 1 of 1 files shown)


gassmoeller:
Oh, I should mention this does not add a test in this PR yet, because it only triggers when something is merged into master (so there is no way to see if it works, before this is merged). However, I have tested this with a modified version in my fork of Rayleigh (https://github.com/gassmoeller/Rayleigh/actions/runs/3053764309) and it produced an image correctly (https://hub.docker.com/repository/docker/gassmoeller/rayleigh).

tukss (approved):
This looks correct to me, but as you mention there is no way to check everything before it's merged. As this is only modifying the docker build, I suggest we merge this now and I'll do tests with the newly built container afterwards.

One question: It looks like this will only build containers for new releases. Do we also want to build containers also for the current and older releases? Or do we have those already?

gassmoeller:
> One question: It looks like this will only build containers for new releases. Do we also want to build containers also for the current and older releases? Or do we have those already?

We do not have them at the moment. I would say let's first update the content of the image (to include Jupyter and the conda environment), then we can consider if we want to build images for older releases. I think it makes most sense getting the current and future versions.

PR #387, opened 2022-09-14
new buildenv using Ubuntu Jammy

tukss:
This creates a new buildenv container using the most recent LTS version of Ubuntu. As packages got updated since Ubuntu Bionic, we can now build Rayleigh and the documentation without relying on pip, except for `nbstripout`, which still does not have a package. This also switches from the standard LAPACK and FFTW to the MKL versions, which are available in the Ubuntu repository.

Other small changes:
- Using `apt-get` instead of `apt` to avoid the warning about the unstable command-line interface
- Remove the repository lists after package install to keep the container size down.

### DIFF
--- a/docker/rayleigh-buildenv-jammy/Dockerfile
+++ b/docker/rayleigh-buildenv-jammy/Dockerfile
@@ -1,0 +1,41 @@
+FROM ubuntu:jammy
+
+RUN apt-get update && \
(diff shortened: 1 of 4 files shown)


gassmoeller (approved):
Very nice! We should update our github workflow to use this container as base too. Do you want to give that a try or should I do it?

tukss:
I'll give it a try just wanted to get this merged first.

PR #390, opened 2022-09-14
Disable execution of jupyter notebook on readthedocs

### LINKED ISSUE #353: Documentation:  Jupyter notebook python output
There are several places in the documentation where the build process is attempting to run Python code that relies on nonexistent files (on the documentation server).  These are typically cells that illustrate plotting examples.  It would be good to adjust this so that the python code is displayed, but not executed, so that the documentation does not appear broken in places.

A prime example can be seen [here](https://rayleigh-documentation.readthedocs.io/en/latest/post_processing/Diagnostic_Plotting.html#II.-Configuring-Your-Python-Environment).

gassmoeller:
This PR fixes #353 by disabling the execution of jupyter notebooks when the documentation is build on readthedocs. We can later figure out a fancy way of getting the necessary data to readthedocs, but for now this is the simplest way to remove the error messages in the documentation.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -49,6 +49,7 @@ templates_path = ['_templates']
 # This pattern also affects html_static_path and html_extra_path.
 exclude_patterns = ['_build', '**.ipynb_checkpoints']
 nbsphinx_allow_errors = True
+nbsphinx_execute = 'never'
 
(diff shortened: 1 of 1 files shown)


tukss (approved):
I agree that's the best course of action until we have a better solution.

PR #391, opened 2022-09-14
switch github workflow to new container using Ubuntu Jammy

tukss:
This updates our workflow to a new container image. The old workflow used an image in @gassmoeller's namespace on Docker Hub. For testing I switched this to an image in my namespace. Should we use the geodynamics namespace in the long run? I've seen that Github also offers a container registry, so that might be another option. Any ideas?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,9 +12,9 @@ env:
 
 jobs:
   test:
-    name: linux-bionic-container-build
+    name: linux-jammy-container-build
     runs-on: [ubuntu-latest]
-    container: gassmoeller/rayleigh-buildenv-bionic
+    container: tuks/rayleigh-buildenv-jammy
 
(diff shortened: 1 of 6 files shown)


gassmoeller (approved):
Nice, let's see if the tester still passes.

Yes we should upload the buildenv into the geodynamics workspace. Using my user is a leftover, but we can use yours until we have made sure the tests still work. I have access to geodynamics and can upload the new image to there.

tukss:
One thing I don't understand yet is how the integration into readthedocs works. I don't see it as a GitHub action. Is that something external triggered by readthedocs?

gassmoeller:
The readthedocs integration works differently, it is a webhook triggered by github that tells readthedocs to trigger a new build and readthedocs has permissions to add the test result to the PR. No github action involved.

PR #392, opened 2022-09-15
Update containers available to users

tukss:
This updates all containers to the new Ubuntu Jammy base. Before we merge this the corresponding base containers (`rayleigh-buildenv-jammy`, `rayleigh-devel-jammy`) should be pushed to the `geodynamics` namespace on Docker Hub.

This also exposes port 8888 so that Jupyter notebooks can be accessed from outside the container with a command like:

```sh
docker run -it --rm -p 8888:8888 geodynamics/rayleigh
jupyter notebook --ip=\*
```

### DIFF
--- a/docker-devel
+++ b/docker-devel
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-docker run -it --rm -v $HOME:/work -e HOSTUID=$(id -u) -e HOSTGID=$(id -g) -e HOSTUSER=$(id -n -u) geodynamics/rayleigh-devel-bionic:latest
+docker run -it --rm -v $HOME:/work -e HOSTUID=$(id -u) -e HOSTGID=$(id -g) -e HOSTUSER=$(id -n -u) geodynamics/rayleigh-devel-jammy:latest
--- a/docker-devel.bat
+++ b/docker-devel.bat
@@ -1,3 +1,3 @@
 :: echo off
 
-docker run -it --rm -v "%USERPROFILE%":/work -e NOUIDWARN=1 geodynamics/rayleigh-devel-bionic:latest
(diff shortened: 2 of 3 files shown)


gassmoeller:
I am in the process of building and uploading the images to docker hub. It will take a while, because the buildenv container alone is around 5GB (because of the inclusion of latex and python libraries).

gassmoeller:
Ok, images are pushed. Do you want to test it?

tukss:
@gassmoeller, in the container you pushed to `geodynamics/rayleigh-buildenv-jammy` the file `/lib/x86_64-linux-gnu/libz.so.1.2.11` has 0 bytes, which causes `sudo` to fail among other things. When I build the container it works fine and the file actually has contents. Maybe something went wrong during the build or push. Can you rebuild and push it?

gassmoeller:
Weird, my local images have the file with non-zero size, and I cannot even push the image again, because it says the online version is up-to-date.  I downloaded the container on another machine and it also appears correctly there. Can you provide a list  of steps how to reproduce the problem? I can try to reproduce it on my system.

tukss:
This is fascinating. It must be related to the docker install on my laptop. When I delete and pull the image again there, the file is still empty. When I do the same thing on my desktop, which hasn't seen the image before, everything works fine and the files have the expected content.

As this seems to be related to my Docker install only, I wouldn't worry about it. I'll test the images on my desktop and report back.

tukss:
The `docker-devel` script still works as expected. As far as I understand, the updated `geodynamics/rayleigh` container will only be pushed once this is merged, so I can only test a locally built version.

gassmoeller:
Let's merge and find out. Since this container is not in large use at the moment anyway there is no harm in trying.

gassmoeller (approved):

PR #393, opened 2022-09-15
Update test workflow

gassmoeller:
This makes use of #391 to simplify some of the mpirun options, which can now be handled by environment variables. Also change the trigger action for the workflow to avoid running the workflow when users push random branches to their personal forks (= waste of energy).


tukss (approved):
These changes all improve our workflow. The container on `geodynamics` works for me and the tests pass.

PR #397, opened 2022-09-15
Reviewed the plotting section

nicholasjnelson:
Note that there is an issue with the menu related to the Jupyter notebook we are importing.


feathern (approved):
This compiles correctly on local machine as well.   We will revisit the 'contents' later today.

PR #398, opened 2022-09-15
Fix path to CIG logo on subpages of documentation.

gassmoeller:
This fixes a little quirk in the current documentation. The CIG logo is only displayed correctly on the homepage, because it is inserted as static html code to every page. On subpages the path is currently not correct.
 In order for this to work the path needs to be absolute to the homepage, not relative to the current page.


feathern (approved):
Great -- docs coming together!

PR #399, opened 2022-09-15
Deploy the new user guide and associated revisions

nicholasjnelson:
We did not delete the old user manual files, but they are not referenced. Information has been copied over to the new files.


feathern (approved):
Wonderful.  This builds well on my local machine.  This is a much-needed re-org of the documentation!

PR #401, opened 2022-09-15
Doc cleanup2

feathern:
This PR completes the initial consistency check of the revised User Guide.  Internal links have been checked and fixed where broken or obsolete.  Various bits of text have been revised to eliminate redundant documentation.   Unused and/or obsolete files from the old user guide have been improved.


nicholasjnelson (approved):
Looks good on local system.

PR #407, opened 2022-09-16
Add installation instructions for Stampede2.

gassmoeller:
This PR changes:
- Adds installation instructions for Stampede2 (I think we should move out all the installation details from Getting Started into a separate Installation section, but that would be something for later).
- Changes User Manual to User Guide (I thought we had settled on User Guide?)
- Removes leftovers from the manual reorganization in index.rst.


feathern (approved):
Looks good, and sorry this took so long.  I got pulled away in the middle of looking at this last week.  And yes, we did agree on 'guide' vs. 'manual.'
-Nick

PR #412, opened 2022-09-27
Update accessing_and_sharing_data.rst

illorenzo7:


feathern:
Ugh.  The build failed due to excessive memory consumption?!   I wonder what changed.  It looks like it failed on a pip command, which has nothing to do with your PR.   I will approve.   I'm curious if the doc build will work after merging.  Can't see why it should, but who knows.
-Nick

feathern (approved):
Approved

PR #414, opened 2022-12-01
Small Section Rewording

feathern:
I think this small change in language is better/clearer.  It feels odd to refer to 'equation sets.'


gassmoeller (approved):

gassmoeller:
Makes sense.

PR #416, opened 2022-12-05
Adding Custom Background dSdr and dTdr via augment_reference.

feathern:
This PR adds the ability for the user to modify the background dSdr and dTdr through the augment_reference routine.  This functionality has been added to allow advection of a background temperature gradient in a Boussinesq model without the need to specify a full custom reference state.   I've added this at the request of @jorafb , so I am assigning the review of this PR to him.

### DIFF
--- a/src/Physics/PDE_Coefficients.F90
+++ b/src/Physics/PDE_Coefficients.F90
@@ -686,6 +686,7 @@ Contains
     Subroutine Augment_Reference()
         Implicit None
         Real*8, Allocatable :: temp_functions(:,:), temp_constants(:)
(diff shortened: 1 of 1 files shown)


illorenzo7 (approved):
Since I've been poking around the custom reference stuff so much, I figured I'd go ahead and approve this (it looks good). But maybe we should give Rafa a chance to look through it before merging?

feathern:
Ok.  Given all the work you're doing in PDE_Coefficients.F90, I think it might be good to go ahead and merge this.  Thanks for having a look.

PR #418, opened 2022-12-12
Description of Streamfunction Approach

feathern:
This PR fills in the streamfunction section of the documentation, which is presently empty.


nicholasjnelson (approved):
Looks good -- it's concise and easy to read.

PR #421, opened 2022-12-13
BUG FIX: Restore_Reference_Defaults and Angular_Velocity

feathern:
The restore_reference_defaults routine was incorrectly setting the default value of angular_velocity to 1.0 instead of -1.0.  This creates a logical error when running benchmark mode alongside a custom reference state.  The result is that the angular_velocity variable is set to 2, regardless of what is specified in the custom_reference file.   Benchmarking mode is intended to work with custom reference states, in part so that this functionality can be validated against known solutions.  The one-line change in PDE_Coefficients.F90 fixes this issue.


illorenzo7 (approved):
Looks good, thanks!

PR #422, opened 2022-12-14
Adding Tests for Custom Reference State Functionality

feathern:
Recently, we inadvertently broke part of the custom-reference-state functionality.   This PR adds two tests.

The first test (named 'basic') runs the case 0 benchmark for 200 timesteps, but specifies the Rayleigh number, Ekman number, etc. completely via the custom-reference interface.   The second test, named 'augment,' uses the with_custom feature to modify the case 0 model such that the Rayleigh number is a bit higher and such that internal heating is used.


feathern:
OK, this seems to be working.   Both of these tests appear to pass and the output in the log file looks correct.  I would like to assign this to @gassmoeller and/or @cianwilson who wrote the existing tests.  Please have a careful look and let me know if I need to change anything.  This is the first test I've written.

src/Physics/PDE_Coefficients.F90
```diff
@@ -1150,7 +1150,7 @@ Subroutine Restore_Reference_Defaults
         poly_mass = 0.0d0
         poly_rho_i = 0.0d0
 
-        Angular_Velocity = 1.0d0
+        Angular_Velocity = -1.0d0
```
gassmoeller, line 1153: Was this a bug you fixed as part of this PR?

tests/custom_reference/augment/main_input
```diff
@@ -0,0 +1,69 @@
+! This is a smaller and faster version of the
+! input_examples/benchmark_diagnostics_input model. The output of the model is
+! not yet in steady-state, but only used as a regression test. Whenever
+! something changes the functionality of tthe code, it will hopefully affect
```
gassmoeller, line 4:
```suggestion
! something changes the functionality of the code, it will hopefully affect
```

gassmoeller (approved):
Looks good, and yes until we set up a test management suite (like ctest) this is the way to add new tests.

If the change I commented on was intended and you fix the typo this is good to merge.

src/Physics/PDE_Coefficients.F90
illorenzo7: I actually had a question about this too. I did all my pending pull requests not realizing PDE_Coefficients was modified by a pending pull request. That was my bad! But, is there any way this change could be removed for now without affecting the tests? I don't know Git well enough to be sure my pending pull requests will have conflicts if one line of the relevant file was changed. If removed now, this bug should also be addressed in #427, 

If not, I'm sure I can patch this one change to modify the other pull requests, so it's not a big deal.

feathern: I'm actually not sure why that's appearing as a change.  I suspect it's because I created this branch before I made a separate branch to initiate PR 421:  https://github.com/geodynamics/Rayleigh/pull/421

This isn't actually a change in this PR.  It's a bug fix that was already implemented, and if @illorenzo7 was working with the latest version, it shouldn't be a problem.  FWIW, he approved that PR!

tests/custom_reference/augment/main_input
feathern: Good with the typo fix too.  Thanks!  I'm going to merge now.

src/Physics/PDE_Coefficients.F90
illorenzo7: Oh, yeah I remember that now! 

Anyway, I rebased my pending pull requests after you just merged and it looks good---they are running through the checks now.

PR #424, opened 2022-12-14
Fix thermal en eqn codes in docs

illorenzo7:
This fixes two minor Latex errors for the codes 1406 and 1436 in the thermal energy equation.


feathern (approved):
Good catch.  Thanks for fixing this!

PR #426, opened 2022-12-24
Tidy up initialization block of the PDE_Coefficients file

illorenzo7:
There should be no functional change to the code -- just added some comments and reordered/regrouped some of the initialization commands.


feathern (approved):
Looks good to me.

PR #427, opened 2022-12-24
Fix errors in Restore_Reference_Defaults() subroutine

illorenzo7:
The Restore_Reference_Defaults() routine of PDE_Coefficients was a bit out of date. Some variables were not restored to their proper values, or at all. Not all the allocatable arrays were deallocated. This pull request should fix these errors. It incorporates the code cleanup from my other pull request (#426 ).

### DIFF
--- a/src/Physics/PDE_Coefficients.F90
+++ b/src/Physics/PDE_Coefficients.F90
@@ -1173,13 +1173,13 @@ Contains
 
         ! Custom reference-state variables (reference_type = 4)
         !       NOTE: n_ra_constants / n_ra_functions do not have default values (but maybe do, if you consider the Allocate_Reference_State() routine)
-        with_custom_reference = .false.
-        override_constants = .false.
-        override_constant(1:max_ra_constants) = .false. ! in namelist
-        with_custom_constants(1:max_ra_constants) = 0   ! in namelist
-        with_custom_functions(1:max_ra_functions) = 0   ! in namelist
-        ra_constants(1:max_ra_constants) = 0.0d0        ! in namelist
-        custom_reference_read = .false.
+        !with_custom_reference = .false.
+        !override_constants = .false.
+        !override_constant(1:max_ra_constants) = .false. ! in namelist
+        !with_custom_constants(1:max_ra_constants) = 0   ! in namelist
+        !with_custom_functions(1:max_ra_functions) = 0   ! in namelist
+        !ra_constants(1:max_ra_constants) = 0.0d0        ! in namelist
+        !custom_reference_read = .false.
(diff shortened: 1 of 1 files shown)


illorenzo7:
Hmmm....these pull requests are now failing after merging @feathern's custom reference tests. Fortunately, this one (which only modifies the "Restore_Reference_Defaults()" routine) is also failing, which means there must be a problem with my fix to the restore defaults---probably specifically related to the custom reference stuff. Let me look into this. Hopefully it's just an issue with "Restore_Reference_Defaults()" and an easy fix.

feathern:
Looks like your fix passed the check.  I'm cooking dinner right now, but I'll review this in the morning and we can go from there.  And yes, the restore_reference_defaults was way out of date.  A conversation for our monthly meeting I think (others are out of date too).

illorenzo7:
Yes, it seems like I can't reference the custom reference stuff in the restore defaults routine yet. I'm not entirely sure what's going on, or whether the problem is on this end, or the new custom tests' end. Anyway, I will try rebasing the other pull requests of this (slightly "incomplete" Restore_Reference_Defaults() routine--should be a relatively easy thing to fix later, and I'd like to get that other stuff (which feels more critical, since there were some bugs) rolling. Hopefully this was the whole issue, and the other pull requests pass the checks after rebasing. We'll see!

feathern:
Two things:
1)  I can merge this once the commented lines are deleted rather than commented.  I'm referring to the lines you had to fix to make this work.
2)  The issue here was that restore_reference_defaults is called as part of the benchmarking initialization.  We wanted to ensure that benchmarking mode could be used as a way of testing the custom reference state framework.  When benchmarking mode is used alongside a custom reference state, a few variables (like ref_type) are saved and restored following the call to restore_reference_defaults.  The lines you commented out were effectively undoing whatever had previously been read in from main_input describing the custom reference state, and so the tests were rightfully failing.

It's worth noting that restore_reference_defaults is now only used for benchmarking and doesn't necessarily need to actually restore all of the default values as it did in the past.

illorenzo7:
OK, I will go ahead and delete these. I'm fairly unfamiliar with these routines. Basically, I went off the lines

1145         ! Restore all variables in this module to their default state.
1146         ! Deallocates all allocatable module variables.

Anyway, it sounds like we'll be discussing these routines in general, so as long as you think it's doing what it's supposed to now, that's fine.

illorenzo7:
Just fixed this but it triggered another test (I submitted commit directly from Github).

feathern (approved):
OK, looks good. Merging now.

PR #429, opened 2022-12-25
Fix bugs for custom diffusions

illorenzo7:
There was an issue in the logic for specifying custom scalar diffusions. Only nu_type, kappa_type, and eta_type were checked to be 3, with the result that if custom scalar diffusions were used, they might not be read from the binary file (unless reference_type was 4, in which case, the custom reference file had already been read). 

Similarly, for reference_type != 4 there was an indexing issue in temp_constants, that led to seg faults if one of the reference_types was 3. @feathern perhaps this is related to the issues you were seeing regarding custom reference and the scalar fields? 

Anyway, this pull request should fix these bugs. It incorporates my other pull requests (#427 and #426). 

Finally, there is an outstanding issue (that is supported by my preliminary testing) with custom diffusions (and again reference_type !=4 ). If one custom diffusion is used (but not all), then temp_functions for the other diffusions will be whatever ra_functions used to be before the initialize_diffusivity calls (likely zero). I think the actual diffusions in Rayleigh will be set correctly (and so will ra_functions), but then ra_functions will get overwritten by the incorrect temp_functions. The code will then run correctly but the output of the equation_coefficients file will be corrupt. Leaving this for now.

### DIFF
--- a/src/Physics/PDE_Coefficients.F90
+++ b/src/Physics/PDE_Coefficients.F90
@@ -1280,21 +1280,51 @@ Contains
         Implicit None
         Integer :: i
         Real*8, Allocatable :: temp_functions(:,:), temp_constants(:)
-        Logical :: restore
+        Logical :: restore, need_custom
 
         restore = .false.
 
         Call Allocate_Transport_Coefficients
 
-        If ((.not. custom_reference_read) .and. &
-            ((nu_type .eq. 3) .or. (kappa_type .eq. 3) .or. (eta_type .eq. 3))) Then
+        ! Figure out if we need to read anything from the custom file 
+        ! (many "types" to check now because of the new scalar diffusion coefficients)
+        need_custom = .false. 
+        If ( (nu_type .eq. 3) .or. (kappa_type .eq. 3) .or. (eta_type .eq. 3) ) Then
+            need_custom = .true.
+        EndIf
+
+        Do i = 1, n_active_scalars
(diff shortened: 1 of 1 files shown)


illorenzo7:
Update: there was another bug in initializing ref%chi_buoyancy_coeff (the first index was zero instead of one). Someone more familiar with the scalar fields (@feathern or @cianwilson) should check on this to make sure everything is going as intended as far as the indexing is concerned. 

My (potentially flawed) interpretation: Say j = 1, 2, 3, ... , n_active_scalars. From what I understand, the user can set the constant multiplying the relevant function via the [11 + (j-1)*2]th constant. They can set the constant multiplying the buoyancy force (f_2) via the [12+(j-1)*2]th constant. This is at least what the code seems to reflect. 

It also seems that the buoyancy scalar diffusion constants are never set for reference_types = 1, 2, 3. Since I'm already thinking about this today, I will fix this in a separate pull request.

illorenzo7:
Great! The same fix for #427 works here. I'll do the others now.

feathern (changes requested):
This looks good save for two things.
1) The conflict ('object' vs. 'derived type') that needs to be resolved.  
2) As we discussed offline, please remove the comment about a bug that needs to be fixed (maybe just paste that into a brief issue note in Github).

feathern (approved):

feathern:
Great -- merging.

PR #430, opened 2022-12-30
[WIP] Reorder Logic in PDE_Coefficients.F90

illorenzo7:
The purpose of this pull request is to leave the current reference state framework entirely unchanged (I am, however, including my pending pull requests, #426, #427, and #429), but re-order the logic in setting each reference type. Essentially, I want ra_constants and ra_functions to be set before the "ref" object that Rayleigh actually uses.

I think this formulation would be better for two reasons: First, future reference types should be easier to implement, since all the coder needs to do is formulate the problem according Rayleigh's "constants and functions" documentation, and then transcribe that problem into a new subroutine. No one will have to worry about the non-documented "ref" object (or "nu" or "kappa", etc.) ever again! Second, I've uncovered at least one instance where the "ref" object is set correctly, but ra_constants/ra_functions are not. I think the new logic should minimize the instances where this can happen, since the ra_constants/ra_functions will necessarily be consistent with "ref" at least at the point "Call Initialize_PDE_Coefficents()" happens.

This is a work in progress, because there's a lot of restructuring that could potentially break things inadvertently. In particular, it would be great if someone more familiar with the "ref" object (probably @feathern) could make sure the new "Initialize_PDE_Coefficients()" subroutine is translating constants/functions correctly. 

I have currently tested basic prototypes for reference_type = 1, 2, 3 (comparing "equation_coefficients" after one time step), plus the Christensen case 0 benchmark, all of which went OK. I will do reference_type = 4 (after which the pull request should be complete, modulo debugging) and re-run all the benchmarks after the whole pull request is done.

### DIFF
--- a/src/Physics/Main.F90
+++ b/src/Physics/Main.F90
@@ -82,13 +82,14 @@ Contains
         Call Initialize_Benchmarking()
 
         Call Initialize_FFts()
+
         Call Initialize_Reference()
+        Call Initialize_Transport_Coefficients()
+        Call Initialize_PDE_Coefficients()
 
         Call Initialize_Field_Structure()
         Call Initialize_Checkpointing()
-        Call Initialize_Transport_Coefficients()
         Call Initialize_Boundary_Conditions()
-
 
         Call Write_Equation_Coefficients_File()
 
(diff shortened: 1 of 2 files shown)


illorenzo7:
Going to close this and start over --- something is failing one of the tests, so I'm going to integrate these changes into a new pull request, piece by piece.

PR #431, opened 2022-12-31
Set ra_constants/ra_functions before "ref" in PDE_Coefficients.F90

illorenzo7:
A replacement version of #430 (now closed). That pull request wasn't passing the "linux-jammy-container-build", so I am integrating the changes slowly (one at a time) to make sure they pass all checks. 

The purpose of this pull request is to leave the current reference state framework entirely unchanged (I am, however, including my pending pull requests, https://github.com/geodynamics/Rayleigh/pull/426, https://github.com/geodynamics/Rayleigh/pull/427, and https://github.com/geodynamics/Rayleigh/pull/429), but re-order the logic in setting each reference type. Essentially, I want ra_constants and ra_functions to be set before the "ref" object that Rayleigh actually uses.

I think this formulation would be better for two reasons: First, future reference types should be easier to implement, since all the coder needs to do is formulate the problem according Rayleigh's "constants and functions" documentation, and then transcribe that problem into a new subroutine. No one will have to worry about the non-documented "ref" object (or "nu" or "kappa", etc.) ever again! Second, I've uncovered at least two instances where the "ref" object is set correctly, but ra_constants/ra_functions are not. I think the new logic should minimize the instances where this can happen, since the ra_constants/ra_functions will necessarily be consistent with "ref" at least at the point "Call Initialize_PDE_Coefficents()" happens.

For the old pull request, I tested basic prototypes for reference_type = 1, 2, 3, 4 and tested the benchmarks (even the MHD Anelastic one!) But something was messing up the continuous integration, so I am taking things more modularly this time around.

### DIFF
--- a/src/Physics/Main.F90
+++ b/src/Physics/Main.F90
@@ -85,6 +85,7 @@ Contains
 
         Call Initialize_Reference()
         Call Initialize_Transport_Coefficients()
+        Call Initialize_PDE_Coefficients()
 
         Call Initialize_Field_Structure()
         Call Initialize_Checkpointing()
--- a/src/Physics/Main.F90
+++ b/src/Physics/Main.F90
@@ -82,13 +82,14 @@ Contains
         Call Initialize_Benchmarking()
 
         Call Initialize_FFts()
+
         Call Initialize_Reference()
+        Call Initialize_Transport_Coefficients()
 
         Call Initialize_Field_Structure()
         Call Initialize_Checkpointing()
-        Call Initialize_Transport_Coefficients()
-        Call Initialize_Boundary_Conditions()
 
+        Call Initialize_Boundary_Conditions()
 
         Call Write_Equation_Coefficients_File()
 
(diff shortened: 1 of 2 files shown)


illorenzo7:
OK, the other pull requests passed the checks except this one. Time to "integrate my changes in slowly" again!

illorenzo7:
OK, I this is finally ready for review (assuming the basic premise that ra_constants/ra_functions being set before "ref" is a good idea). Note that this pull request should basically leave the code functionality unchanged, but it does fix the bug wherein the equation_coefficients file is corrupt when one diffusion time is custom and the other isn't (which occurs for reference_type = 1, 2, 3).

illorenzo7:
Also, the pull requests weren't quite sequential, so this should probably be reviewed AFTER #432. Although the commits are all the same if this one gets reviewed first, then probably #432 becomes irrelevant.

illorenzo7:
I am going to close both this request and #432, and start fresh. Enough has changed that merging can't be done automatically. I also know pretty much what was causing problems before, so I can do this request (including the changes implemented in #432) as only one or two commits.

PR #432, opened 2022-12-31
Initialize buoyancy coefficients for the scalar diffusions (reference_type = 1, 2, 3)

illorenzo7:
The constants associated with the buoyancy forcing for the active scalars (I believe c_[12 + (j-1)*2], where j runs from 1 to n_active_scalars) were never initialized for the default reference types. 

Hopefully they are initialized properly now. This pull request incorporates my previous #426, #427, and #429.

PR #434, opened 2023-01-06
Set ra_constants/ra_functions before "ref" (fewer commits)

illorenzo7:
New version of #431 (and includes #432) which is now closed. This is not ready yet (still doing one commit at a time). I will remove this line once it's ready. 

The purpose of this pull request is to leave the current reference state framework almost completely unchanged, but re-order the logic in setting each reference type. Essentially, I want ra_constants and ra_functions to be set before the "ref" object that Rayleigh actually uses.

I think this formulation would be better for two reasons: First, future reference types should be easier to implement, since all the coder needs to do is formulate the problem according Rayleigh's "constants and functions" documentation, and then transcribe that problem into a new subroutine. No one will have to worry about the non-documented "ref" object (or "nu" or "kappa", etc.) ever again! Second, I've uncovered at least two instances where the "ref" object is set correctly, but ra_constants/ra_functions are not. I think the new logic should minimize the instances where this can happen (as well as fix the two bugs just mentioned). This is because the ra_constants/ra_functions will necessarily be consistent with "ref" at least at the point "Call Initialize_PDE_Coefficents()" happens.


illorenzo7:
OK, @feathern I am realizing that this pull request is (1) too big to do all at once because everything has to be perfect for it to work (and it makes the diffs awful for you, who will likely review it) and (2) tricky to do piece by piece, with everything working at each step. What I am going to do is close this (yet again) and make a series of pull requests that all follow each other, and pass the tests at each stage. 

That should be easier on the reviewer (as far as the diffs go, and following how everything is working) and also harder for me to inadvertently break things, since I keep doing that when trying to rebase/edit whole files!

PR #435, opened 2023-01-06
move Call Initialize_Transport_Coefficients() further up

illorenzo7:
This moves the call to Initialize_Transport_Coefficients() further up (to directly follow Initialize_Reference_State()) in Main_Initialization().

### DIFF
--- a/src/Physics/Main.F90
+++ b/src/Physics/Main.F90
@@ -82,13 +82,14 @@ Contains
         Call Initialize_Benchmarking()
 
         Call Initialize_FFts()
+
         Call Initialize_Reference()
+        Call Initialize_Transport_Coefficients()
 
         Call Initialize_Field_Structure()
         Call Initialize_Checkpointing()
-        Call Initialize_Transport_Coefficients()
-        Call Initialize_Boundary_Conditions()
 
+        Call Initialize_Boundary_Conditions()
 
         Call Write_Equation_Coefficients_File()
 


feathern:
I think we should talk at the next developer meeting about removing the use of the reference structure in the initialization routines.  I take it that's what these various PRs are working toward.  Let's hold off on merging these PRs for now.  Happy to talk offline sometime next week too.  I worry about this impacts the readability/understandability of the code.

illorenzo7:
OK, that sounds fine. I also fully realize that I went overboard on making pull requests without having the requisite discussion first! Originally, I didn't think this change would be so big. Anyway, live and learn (to not do major pull requests over winter break haha). 

Just two things I want to reiterate before I forget. 

The first main reason for these changes is precisely a readability issue (that might be just from my perspective, so we should certainly discuss). I've always found it difficult to understand/edit the PDE_Coefficients file because of the undocumented "ref" object. Usually I do things by first writing down how each part of "ref" corresponds to the c's and f's, which I infer from the Read_Custom_Reference() routine and am never fully sure that I've gotten right. These changes were my way of removing "ref" entirely from the user/coder's experience so I'd never have to worry about it again! If these pull requests aren't desirable, I might also be just as happy if the "ref" object and its relation to the c's and f's were explicitly documented (and checked for accuracy  by you). 

The second issue is I think a bit more critical. Currently there is no guarantee that in non-standard situations (i.e., involving custom reference stuff or the new scalar fields) that the equation_coefficients file will accurately reflect the equations the code is actually solving. I believe I've uncovered at least two cases where this happens and I wonder if there are more. Of course, the way I've formulated the fix here is not the only solution (we could, for example, never modify "ref" or "nu", etc. without also modifying ra_constants/ra_functions in parallel; that could leave the current structure of the code largely intact).

PR #442, opened 2023-01-30
Initialize scalar buoyancy coeffs2

illorenzo7:
Adds loops to Constant_Reference(), Polytropic_ReferenceND(), and Polytropic_Reference() to initialize the previously zeroed-out active-scalar buoyancy constants.


feathern (approved):
This looks good.  Note that we need to revisit the buoyancy terms related to active scalar fields when they're used with a dimensional polytropic reference.

Issue #433, opened 2023-01-06
Possible bug in equation_coefficients file


illorenzo7:
I believe there is a bug in the output of the equation_coefficients file when playing around with custom diffusion types. See the routine Initialize_Transport_Coefficients() in PDE_Coefficients.F90. If one diffusion type is custom but another isn't, then the ra_constants/ra_functions associated with the non-custom diffusion will be set correctly by the Initialize_Diffusivity routine, but then overwritten, likely with erroneous values, when they are "restored."       

I believe this issue will be fixed as part of #431, assuming that pull request goes through.

illorenzo7:
This should have been addressed with the recent pull request #443. Closing.

PR #443, opened 2023-01-30
Fix equation coefficients custom diffusions

illorenzo7:
This pull request "safeguards" the diffusion coefficients part of the equation_coefficients file. It adds the new subroutine Set_Diffusivity_Equation_Coefficients(), called at the end of Initialize_Transport_Coefficients(), to ensure that equation_coefficients always reflects what the code actually solved, at least for the diffusion pieces. 

This should fix the bug mentioned in #433 and guard against similar possible bugs in the future.


feathern (approved):
This also looks good to me.

PR #445, opened 2023-02-12
fixed bug in initialization of T_c for dimensional anelastic

illorenzo7:
This is similar in spirit from my #212 from a few years ago. 

There was a bug in the definition of T_c for reference_type = 2 (dimensional anelastic mode). I believe T_c was being calculated from the ideal gas law assuming 

gas_constant = c_p / (n + 1)

That is only true, however, for adiabatic gases (n = 1.5), not for other n's. In general it is 

gas_constant = (gamma - 1) * c_p / gamma

where gamma = 5/3.

### DIFF
--- a/src/Physics/PDE_Coefficients.F90
+++ b/src/Physics/PDE_Coefficients.F90
@@ -521,7 +521,7 @@ Contains
         Integer :: i
         Real*8 :: zeta_0,  c0, c1, d
         Real*8 :: rho_c, P_c, T_c,denom
-        Real*8 :: thermo_gamma, volume_specific_heat
+        Real*8 :: thermo_gamma, volume_specific_heat, gas_constant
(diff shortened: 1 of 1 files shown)


illorenzo7:
Update: @feathern realized that as it stands, this pull request would break the non-dimensional form of the Jones benchmark, which uses n = 2. Effectively Jones et al. (2011) treat the non-ideal Jovian atmosphere as ideal, with n = 2 and gamma = 3/2, which is conventional practice to approximate the molecular gas phases near the surface and the metallic hydrogen deeper down. 

For the future, a version of this pull request may make sense, as long as gamma is an input parameter to the code (it will be after #450 is implemented). In that case, the non-dimensional Jones Benchmark could then set gamma = 3/2 and should pass.

PR #446, opened 2023-02-12
fixed bug in initializing aspect_ratio (non-dimensional anelastic)

illorenzo7:
In reference_type = 3 (non-dimensional anelastic), the aspect_ratio (if it was unspecified, i.e., needing to be calculated from rmin and rmax) was incorrectly initialized as rmax/rmin, the reciprocal of what it's supposed to be.

### DIFF
--- a/src/Physics/PDE_Coefficients.F90
+++ b/src/Physics/PDE_Coefficients.F90
@@ -432,7 +432,7 @@ Contains
         Endif
 
         If (aspect_ratio .lt. 0) Then
-            aspect_ratio = rmax/rmin
+            aspect_ratio = rmin/rmax
         Endif
         Allocate(dtmparr(1:N_R), gravity(1:N_R))
         dtmparr(:) = 0.0d0


feathern (approved):
Thanks Loren.  This is a good catch.  I was just double-checking the main_input files from the Heimpel et al. paper.  It looks like the aspect ratio was specified explicitly in the main_input, so that's probably why we haven't noticed this bug yet.

ljhwang:
Is the Heimpel paper mentioned: Heimpel, M.H., and Rakesh, K.Y., and Featherstone, N.A., and Aurnou J.M., Polar and mid-latitude vortices and zonal flows on Jupiter and Saturn, Icarus, 379, 114942, Doi:https://doi.org/10.1016/j.icarus.2022.114942, 2022. ?

feathern:
Yep, that's the one.  The models that Moritz pointed people to are the ones published in that paper, so it shouldn't have an impact.

PR #447, opened 2023-02-14
Safeguard equation_coefficients file to reflect the actual code

illorenzo7:
This pull requests adds a new subroutine,

Set_Reference_Equation_Coefficients()

to the PDE_Coefficients module, which is called at the end of the Initialize_Reference() subroutine. The subroutine sets ra_constants and ra_functions (except ones having to do with diffusion, heating, and buoyancy) directly from the "ref" object and is independent of reference_type. This reduces code bloat (previously there were several repeated lines at the end of each ...Reference() subroutine specifying ra_constants/ra_functions in the exact same way), should ensure that the equation_coefficients file always reflects the equations the code actually solved, and should make future reference_type's less error-prone.

This pull request is the natural follow on to #443, which safeguarded the ra_constants/ra_functions having to do with the diffusions.

### DIFF
--- a/src/Physics/PDE_Coefficients.F90
+++ b/src/Physics/PDE_Coefficients.F90
@@ -1537,8 +1537,7 @@ Contains
         ! Set the equation coefficients (apart from the ones having to do with diffusivities / heating)
         ! for proper output to the equation_coefficients file    
         
-        ! All of these have a reference-type-independent correspondence to the "ref" object
-        ! except for the following buoyancy constants
+        ! The buoyancy constants are set dependent on each reference type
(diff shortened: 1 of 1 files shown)


feathern (approved):
Sorry this took so long.  I think I reviewed this a while back and somehow forgot to hit 'OK' and merge.   I just went through again and this looks good.   I do note that viscous_amp and ohmic_amp are nonzero for the constant reference initialization, despite the fact that viscous and ohmic heating are explicitly turned off.  These should probably be set to zero for consistency, but that can be done in another PR -- just noting it here.

PR #449, opened 2023-03-07
Add a link to the INCITE dataset.

gassmoeller:
Add the INCITE dataset to the list of accessible Rayleigh data. We are in the process of finishing up the website allowing access to that data. A probably final version (not yet merged) can be seen here: https://incite--13.org.readthedocs.build/en/13/.


feathern (approved):
Looks good.   We need to talk to Philipp about this test that's failing for every PR -- ignoring that for now.  I think a reference file changed somewhere and needs to be updated.

PR #450, opened 2023-03-17
New general anelastic non-dimensional subroutine (reference_type = 5)

illorenzo7:
Implements to subroutine

Polytropic_Reference_ND_General()

in PDE_Coefficients. Note that my other pull request (#447) is included here. This new functionality allows user to specify simulation using Prandtl number, Rayleigh number (flux or otherwise), Ekman  number, magnetic Prandtl number, and four number characterizing the polytrop ice: specific-heat ratio (default 5/3), poly tropic index, number of density scale heights, and shell aspect ratio. 

User can choose between non-dimensionalizing the reference state using the profile values at the inner radius, outer radius, or volume-averaged over the shell (default). 

User can non-dimensionalize time using either the rotation rate (time = 1/Omega_0) or viscous diffusion time. 

User can non-dimensionalize the length scale using an arbitrary value, with default being the shell depth.

### DIFF
--- a/src/Physics/PDE_Coefficients.F90
+++ b/src/Physics/PDE_Coefficients.F90
@@ -514,7 +514,7 @@ Contains
     Subroutine Polytropic_ReferenceND_General()
         Implicit None
         Integer :: i
-        Real*8 :: c0, c1, poly_n_ad, numer, denom1, denom2, norm, tol
+        Real*8 :: c0, c1, poly_n_ad, norm, tol
         Real*8, Allocatable :: gravity(:), flux_nonrad(:), partial_heating(:), nsquared(:), &
             & zeta(:), dzeta(:), d2zeta(:), dlnzeta(:), d2lnzeta(:)
         Character*12 :: dstring
@@ -748,11 +748,12 @@ Contains
             norm = four_pi*norm/shell_volume
             nsquared = nsquared/norm
 
-            Dissipation_Number = (poly_n + 1.0d0)/(poly_n_ad + 1.0d0)
-            numer = 3.0d0*aspect_ratio*(1.0d0 - aspect_ratio)**2 * (1 - exp(-poly_Nrho/poly_n))
-            denom1 = (3.0d0*aspect_ratio/2.0d0) * (1.0 - aspect_ratio**2) * (1.0d0 - exp(-poly_Nrho/poly_n))
-            denom2 = -(1.0d0 - aspect_ratio**3)*(aspect_ratio - exp(-poly_Nrho/poly_n))
-            Dissipation_Number = Dissipation_Number * numer / (denom1 + denom2)
+            Call Integrate_in_radius(nsquared,norm)
+            norm = four_pi*norm/shell_volume
+            nsquared = nsquared/norm
+
+            ! ref%temperature(N_R) is now T_inner/volav(T); gravity(N_R) = g_inner/volav(g)
+            Dissipation_Number = Dissipation_Number*ref%temperature(N_R)/gravity(N_R)
         Endif 
 
         ! This was all assuming the H in Dissipation_Number was the shell_depth. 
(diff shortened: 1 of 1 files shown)


src/Physics/PDE_Coefficients.F90
```diff
@@ -491,12 +508,329 @@ Subroutine Polytropic_ReferenceND()
...
+        ! User can specify their own custom Length_Scale,
+        ! different than the shell depth
+        ! But by default, use the shell depth
+        If (Length_Scale .lt. 0) Then ! This is default
+            Length_Scale = shell_depth
+            ND_Length_Shell_Depth = .true.
+        Endif
+
+        ! Determine how user wants to specify Ra, Chi_A_Ra, and B_visc.
+        ! (The modified versions or not).
```
feathern, line 540: We should have a branch for a model without rotation.

src/Physics/PDE_Coefficients.F90
```diff
@@ -491,12 +508,329 @@ Subroutine Polytropic_ReferenceND()
...
+
+        ref%dlnrho = poly_n*dlnzeta
+        ref%dlnT = dlnzeta
+        ref%d2lnrho = poly_n*d2lnzeta
+
+        gravity = (rmin**2)*OneOverRSquared
+
+        denom1 = (1.0d0 - exp(-poly_Nrho/poly_n))/(1.d0 - Aspect_Ratio)
+        denom2 = -(Aspect_Ratio - exp(-poly_Nrho/poly_n))/(Aspect_Ratio*Shell_Depth)
+        nsquared = (rmin/radius)**3 / (denom1 + denom2*radius)
```
feathern, line 690: Could we explicily write dsdr in terms of dlnrho and dlnP or dlnT and then multiply by the appropriate amplitude control parameter?

src/Physics/PDE_Coefficients.F90
```diff
@@ -491,12 +508,329 @@ Subroutine Polytropic_ReferenceND()
...
+            ref%temperature = ref%temperature/norm
+
+            Call Integrate_in_radius(gravity,norm)
+            norm = four_pi*norm/shell_volume
+            gravity = gravity/norm
+
+            Call Integrate_in_radius(nsquared,norm)
+            norm = four_pi*norm/shell_volume
+            nsquared = nsquared/norm
+
```
feathern, line 729: Can you compute dissipation number in a similar way to the new way of calculating dsdr?

src/Physics/PDE_Coefficients.F90
```diff
@@ -491,12 +508,329 @@ Subroutine Polytropic_ReferenceND()
...
+            Adiabatic_Polytrope = .true.
+        Else
+            Adiabatic_Polytrope = .false.
+        Endif
+
+        ! Start printing information about the reference state
+        If (my_rank .eq. 0) Then
+            Call stdout%print(" ---- Reference type           : "//trim(" Polytrope (Generalized Non-dimensional)"))
+        Endif
+
```
feathern, line 580: Suggest output warnings, but don't actually change buoyancy frequency.

src/Physics/PDE_Coefficients.F90
illorenzo7: Actually I think now it's fine. At most, we could have the code throw a warning if they set rotation = .false but decide to set a non-zero Modified_Rayleigh_Number (say). But we don't do anything like that for e.g. the Magnetic_Prandtl_Number (the user can set that with magnetism=.false. and no warning will be thrown).

src/Physics/PDE_Coefficients.F90
illorenzo7: I think what's in there now is the cleanest way to do it (no need to integrate temperature and gravity a second time).

feathern (approved):
Looks good.  Thanks for going over this with me!

PR #451, opened 2023-03-29
remove newline to remove black linter warnings

tukss:
This should fix the CI issues we had. It's probably due to a change in the way `black` formats things. For the record, you can just run `black rayleigh.py` and commit the changes to fix the issue.


tukss:
This should fix the problems noticed in #449.

gassmoeller (approved):

PR #452, opened 2023-04-25
Bug Fix: Custom reference filename being reset

feathern:
There is a logical issue in Benchmarking.F90 that is causing the name of the custom reference file (and also heating_type) to be lost when with_custom_reference is true but no benchmarking mode is active.  This PR fixes the logic.


illorenzo7 (approved):
Looks good @feathern. I'll merge as soon as the check completes.

PR #457, opened 2023-06-12
Fix some references to the master branch

gassmoeller:
The automatic docker build stopped working, because I forgot to update the reference to the master branch. I found one other place where this is referenced as well. Should be all fixed now.


feathern (approved):
Thanks.  Good catch!

PR #461, opened 2023-06-13
Add CITATION file

ljhwang:
If CITATION is present, GitHub creates the drop down 'Cite this repository' under the About section of the repo. The 'View citation file' button displays the CITATION file.

Note that CITATION.cff has a message field that the documentation indicates will modify the default message. However, I did not manage to get this to work. If the project uses the webhook integration between Zenodo and GitHUB, converting to .cff might be useful as it overrides metadata defaults (and will override .zenodo.json).


gassmoeller (approved):
Looks good, thanks.

PR #462, opened 2023-06-14
Fix conda environment installation

gassmoeller:
Our current conda environment seems to fail to resolve on linux machines. It works if I remove pandoc from the environment file. Conda will still be pulled in implicitly, but with a version that seems to satisfy requirements more easily. I can still build the documentation locally, let's see what the online tester says.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -88,6 +88,7 @@ jobs:
       with:
         activate-environment: radev
         environment-file: environment.yml
+        channel-priority: strict
     - name: Build documentation
       continue-on-error: true
       run: |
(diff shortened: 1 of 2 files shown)


gassmoeller:
Ok, so it turns out it was not pandoc after all. During my testing I had set the `channel-priority` setting of conda to `strict`, which means once conda finds a package in one channel (e.g. conda-forge) then it doesnt look for other channels (like defaults) anymore. This removes a lot of possibilities for different package combinations and avoids the problems resolving the environment.
If the testers pass and we agree this is a good solution this is ready to merge.

gassmoeller:
I think this is similar to the time it took before it broke, so yes, I think this it correct.
We could probably do faster with mamba or using mamba's solver in conda, but part of the test is to see if the instructions that we give in the documentation actually work, so I want to keep the two consistent.

feathern (approved):
Great -- hopefully this fix will help on the user side too!

PR #463, opened 2023-06-14
Remove images and references to Ubuntu Bionic

gassmoeller:
Ubuntu 18.04 (Bionic Beaver) is no longer supported. Remove all container image recipes and references and replace with references to Ubuntu 22.04 (Jammy Jellyfish) instead.


gassmoeller:
Rebased, this should be good to go if the testers are happy.

PR #465, opened 2023-06-14
Fix geodynamics/rayleigh docker image for apptainer

gassmoeller:
Our current docker container `geodynamics/rayleigh` includes a user with limited permissions and a home directory that is only readable by that user. This is a problem if the container is run using `singularity` or `apptainer`, because these container systems automatically set the user identity inside the container to the user on the outside. This means the Rayleigh installation in the current container is not usable when using apptainer or singularity. A common solution to this problem is to have the docker container default to the root user.
This way the two use cases (docker / apptainer) both work:
- for docker the user id switched to root inside the container and allows everything, including to install new software if necessary (=better than the current container)
- for apptainer/singularity the user id will be taken from the host system, but because rayleigh is installed in a generally readable location they can execute it normally. In addition, because apptainer/singularity automatically mount the current directory into the container the user is also allowed to write in this directory.

@tukss and @cianwilson I have already rebuild and updated the geodynamics/rayleigh container on docker hub. Could you try to execute that container using Singularity or Apptainer on the systems you have available? I tried on my laptop and one other system and it seems to work.

```
apptainer pull docker://geodynamics/rayleigh
# This should always work
apptainer exec rayleigh_latest.sif mpirun -n 4 rayleigh.opt
# this should only work if MPI inside and outside of container are compatible
mpirun -n 4 apptainer exec rayleigh_latest.sif rayleigh.opt```
```

### DIFF
--- a/doc/source/User_Guide/getting_started.rst
+++ b/doc/source/User_Guide/getting_started.rst
@@ -117,6 +117,7 @@ Docker Container
 ~~~~~~~~~~~~~~~
 
 Docker provides a standardized way to build, distribute and run containerized environments on Linux, macOS, and Windows. To get started you should install Docker on your system following the instructions from `here <https://www.docker.com/get-started>`_. On Linux you can likely also install it from a distribution package (e.g., ``docker-io`` on Debian/Ubuntu). `Podman <https://podman.io/>`_ is an alternative runtime that can run Docker containers and can be used as a drop-in replacement for Docker.
+Singularity/Apptainer are other available alternatives that are more commonly used on HPC systems.
 
 Launching the container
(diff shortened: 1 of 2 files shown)


cianwilson:
Tested this on:
* singularity on a linux box
* docker on a linux box
* docker on an intel mac
* docker on an m1 mac (not built for this but works)
and looks good on all.

Would it be worth adding a line about this being compatible with singualrity/apptainer to the docs?

gassmoeller:
Good point, I know added a line into the docs.

feathern (approved):
This has been sitting unattended for a very long time now!  Since Philipp's tests ran well and the docs were updated, I'm going to approve and merge.

PR #468, opened 2023-06-15
add support for podman

tukss:
This makes the docker-devel script aware of podman and automatically selects it as a fallback. The bat file is not updated as there is no podman on Windows.


gassmoeller (approved):
Looks good.

PR #473, opened 2023-06-16
Including Jupyter notebooks for Ultraspherical polynomials

kyle-augustson:
This commit adds two notebooks that provide some details on Ultraspherical polynomials and how to construct sparse spectral operators using them.  The two examples are close to the structure used in Rayleigh in the sense that they solve initial value problems with one time and one space dimension using pseudo-spectral methods.


feathern:
Closing so a separate branch can be created before review.

PR #476, opened 2023-06-16
This commit adds two jupyter notebooks exhibiting the Ultraspherical polynomials.

kyle-augustson:
The file Scalar_Diffusion.ipnb has some mathematical details about the Ultraspherical polynomials regarding how to build sparse spectral representations of operators such as derivatives.  It also details how to maintain sparsity of those operators even if they are multiplied by non-constant coefficients. The example problem is the 1D diffusion equation both with and without a spatially variable diffusion coefficient for scalar quantity q. This uses the second order Crank-Nicholson method for time stepping. The file Reaction_Diffusion.ipynb shows how to construct a non-linear PDE using the Newell–Whitehead-Segel equation as an example.


feathern (approved):
Awesome.  This should serve as the basis for some solid documentation once we get this formalism implemented into the main code.

PR #477, opened 2023-06-16
Push to ghcr and build for Apple Silicon

gassmoeller:
This PR makes two changes to building the rayleigh docker container:
1. It builds a multi architecture image for amd64 (all usual systems) and arm64 (closer to current Apple systems). Using docker will automatically download the best image for the current system.
2. It pushes the image not just to docker hub, but also to Github Container Registry (ghcr.io). This just provides the image in another location. Docker Hub has been a bit unreliable in the past and we are adding GHCR as back up for CIG packages.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -23,8 +23,18 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v3
 
+      - name: Set up QEMU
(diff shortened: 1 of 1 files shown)


gassmoeller:
This addresses one of the limitations of #465.

gassmoeller:
@cianwilson after this is merged, can you try the docker image again with your M1 machine and check if the speed improves?

PR #478, opened 2023-06-16
Report the diff between expected and actual test report

gassmoeller:
I noticed that our CI does store the diff between expected and actual test results, but it never uploads them to the github page. This change makes it so that users can download the diff file to investigate exactly what went wrong in the test.


feathern (approved):
I can see this being very useful (could have used a diff log a few times myself).  Approved.

PR #480, opened 2023-06-21
put c_8 into mm and pp enth_flux, dt_by_dp

illorenzo7:
I believe the current calculation of enthalpy flux is inconsistent between the total flux and Reynolds decomposition (there is a missing c_8 in definition of dt_by_dp). I think it only affects non-dimensional cases, not the dimensional one (since they're c_8 = 1). This pull request should fix things.


feathern (approved):
Good catch.  This got me thinking about how some of the other energy-equation-related outputs are scaled in a nondimensional, anelastic setup.  I see now that you have issue related to this.  We should definitely give some thought to this.  In any event, this PR is approved -- thanks!

PR #482, opened 2023-06-23
added dim/nondim input file pair

illorenzo7:
@feathern OK I have convinced myself that reference_type = 5 produces the same simulation as reference_type = 2, just rescaled. I added two input files to show how this works. One is the original "model_15" from our parameter survey and the other is the non-dimensional equivalent. 

Note the different conventions in Rossby number. 

I checked volume averaged output non-d numbers, differential rotation, meridional circulation, and radial energy flux balance and they all seem the same (after rescaling) up to < 1%. 

I think the 1%  difference comes from the heating being normalized slightly differently for ref type 5 (it doesn't get rescaled in Transport_Dependencies for ref type 5 to match the conductive flux loss out the boundaries). This makes the input Rayleigh numbers slightly different. 

Anyway, I think we're good to go! And these input files should show people how to use the new reference type.

### DIFF
--- a/input_examples/anelastic_dimensional
+++ b/input_examples/anelastic_dimensional
@@ -1,8 +1,8 @@
 ! This initializes a DIMENSIONAL anelastic reference state (reference_type = 2)
 ! that represents a rotating, solar-like case, spanning the innermost
 ! 3 density scaleheights of the convection zone.
-! This is "model 15" (Ro_c = 0.5, Ra = 10^4, Pr = 1) 
-! from the Boulder Stellar Convection Team's large parameter survey
+! This model has Ro_c = 0.25, Ra = 10^4, Pr = 1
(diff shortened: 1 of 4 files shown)


feathern (changes requested):
Hi Loren,
This looks great, but I was wondering if you could change the filenames and the header text so that we don't mention 'model 15' or the 'Boulder survey.'  Since that model isn't published yet, and since it's number might even change in the publication, it's probably best just to call these files 'anelastic_nondimensional' and 'anelastic_dimensional' or something like that.   Could you also reference the other file in the header?   Thanks!  I think this should just take a sec to change, and then I can approve/merge.

illorenzo7:
@feathern Ok it should be good to go now.

feathern (approved):

feathern:
Looks great - merging!

PR #483, opened 2023-06-27
Moved old spectral utils notebook to legacy folder.

feathern:
This PR moves spectral_operations.ipynb down one level into the 'post_processing/legacy' folder.   The information contained in this notebook should be superseded by that contained in the new spectral_utils_tutorial.ipynb notebook.  Since it's possibly useful to keep around in the near term, I'm moving it to the 'legacy' folder to avoid confusion.


feathern:
This is minor enough that I'm just going to go ahead and merge...

PR #486, opened 2023-06-29
Adding CI Test for Finite-Difference Scheme (uniform grid)

feathern:
This adds an initial test for the new FD functionality when applied to a uniform grid.  It is otherwise the same test and input file used for Chebyshev mode.


illorenzo7 (approved):
Looks good, thanks @feathern

PR #488, opened 2023-10-06
Initial conditions using checkpoint files added to user supplied files

bhishekmanek:
The default initial conditions type in Rayleigh allow reading from a checkpoint file (init_type = -1 for thermal and velocity fields, magnetic_init_type = -2 for magnetic fields). Modifications made in this request allow for new IC's:

1.  init_type = -2 reads checkpoint file as well as a user-supplied entropy profile. The sum total of these two is the new IC. 
2. magnetic_init_type = -2 reads checkpoint file and user-supplied profiles of toroidal and poloidal field components. The sum total of checkpoint file and user-supplied profiles is the new IC. 

An additional subroutine Get_RHS is added to src/Math_Layer/Linear_Solve.F90.

### DIFF
--- a/src/Math_Layer/Linear_Solve.F90
+++ b/src/Math_Layer/Linear_Solve.F90
@@ -641,6 +641,29 @@ Module Linear_Solve
         Endif
     End Subroutine Set_RHS
 
+    Subroutine Get_RHS(eqid,get_from)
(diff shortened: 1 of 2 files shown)


feathern (approved):
Looks good.   I'm going to approve and then submit a small PR of my own removing the !Nick comment.

PR #489, opened 2023-11-25
ref%dsdr should now be scalable by constant c_11

illorenzo7:
All terms in the PDEs have constants out front that can scale the terms, except for the reference entropy advection term f_14. This pull request would insert a new constant (c_11) that can scale ref%dsdr in the custom framework and takes on an appropriate value for the other reference_type's.

### DIFF
--- a/doc/source/User_Guide/physics_math_overview.rst
+++ b/doc/source/User_Guide/physics_math_overview.rst
@@ -147,7 +147,7 @@ respectively. The evolution of :math:`\Theta` is described by
 .. math::
    :label: theta_evol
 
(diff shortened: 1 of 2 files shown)


illorenzo7:
OK, c_11 should now be properly included in documentation. I also noticed nonzero dS/dr was still present for reference_type 3 (but it gets set to zero there) so I corrected that as well. 

I realize I never put the general anelastic mode in the doc, I will do that in a separate pull request.

feathern:
Thank's for updating the documentation.  I'm almost good with this now, but have one small request.  Can you remove the two debugging statements "got here" and "actually got here"?  After that, I'll approve and merge.
-Nick

feathern:
Actually, I'll do that in a separate PR momentarily just for expediency.

feathern (approved):
Approved

PR #490, opened 2023-11-30
guarded against dividing by N^2=0

illorenzo7:
For reference_type = 5, when the buoyancy frequency N^2 was rescaled, this was causing problems with division by 0. I guard against this now with if-statements.


illorenzo7:
Hmm, it looks like my original code had the duplication, which was a mistake on my part, as far as I can tell. I must have removed the duplicate block when I surrounded it by the if-statement.  I don't think removing the duplicate block should affect anything.

feathern (approved):
Looks good.  Approved.

illorenzo7:
Thanks @feathern ! Good catch, I forgot I put those debugging statements there.

PR #492, opened 2023-11-30
Fix with_custom for heating

illorenzo7:
The with_custom framework successfully modifies the heating by a custom reference file. However, the flag "adjust_reference_heating" has been set to .true. earlier (for any heating_type > 0) and so the user's choices will possibly be overwritten in the boundary conditions file. 

This pull request changes adjust_reference_heating back to .false. if the user sets a custom heating profile using the with_custom framework.


feathern:
I'll try to look at this in detail soonish.  Might be a couple of days, but I'm aware of the PR.

illorenzo7:
Because of the latest approved pull requests, I removed the duplicate commit from #490 (the divide by N^2=0 pull request) and then rebased this off "main" to make the analysis of the specific change here more easily readable.

feathern (commented):
This is a good catch of  a very subtle logical issue!  Merging.

feathern (approved):

PR #494, opened 2023-12-06
user can now set any two of Ra, Ra*, and Ek in Reference_Type 5

illorenzo7:
If we disregard Pr, only two of the parameters Ek, Ra, and Ra* (or equivalently Ro_c = sqrt(Ra*)) are independent in a rotating system. This pull request updates the logic so that if two of Ek, Ra, or Ra* are set, the third is automatically set according to

Ra* = Ra Ek^2/Pr

The logic does this by setting all these input parameters to -1 and checking if they are specified (which they will be if they are > 0).

### DIFF
--- a/src/Physics/PDE_Coefficients.F90
+++ b/src/Physics/PDE_Coefficients.F90
@@ -81,8 +81,8 @@ Module PDE_Coefficients
     Integer :: reference_type = 1 
 
     ! Nondimensional variables (reference_type = 1,3)
-    Real*8 :: Rayleigh_Number         = 1.0d0
-    Real*8 :: Ekman_Number            = 1.0d0
+    Real*8 :: Rayleigh_Number         = -1.0d0
+    Real*8 :: Ekman_Number            = -1.0d0
     Real*8 :: Prandtl_Number          = 1.0d0
     Real*8 :: Magnetic_Prandtl_Number = 1.0d0
     Real*8 :: gravity_power           = 0.0d0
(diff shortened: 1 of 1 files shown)


aurnou:
Just wondering, physically, why disregard Prandtl? On Dec 5, 2023, at 6:15 PM, Loren Matilsky ***@***.***> wrote:﻿If we disregard Pr, only two of the parameters Ek, Ra, and Ra* (or equivalently Ro_c = sqrt(Ra*)) are independent in a rotating system. This pull request updates the logic so that if two of Ek, Ra, or Ra* are set, the third is automatically set according to
Ra* = Ra Ek^2/Pr
The logic does this by setting all these input parameters to -1 and checking if they are specified (which they will be if they are > 0).



You can view, comment on, or merge this pull request online at:
  https://github.com/geodynamics/Rayleigh/pull/494

Commit Summary

  73af754  user can now set any two of Ra, Ra*, and Ek in Reference_Type 5


File Changes  (1 file)

  
    M
    src/Physics/PDE_Coefficients.F90
    (68)
  


Patch Links:

  https://github.com/geodynamics/Rayleigh/pull/494.patch
  https://github.com/geodynamics/Rayleigh/pull/494.diff

illorenzo7:
@aurnou, Oh well we don't have to. But my sense is people typically want to set Prandtl to something specific and then play around with the other parameters. It seems unlikely people would want Ra, Ra*, and Ek to all be something specific (thus needing a weird Prandtl number).

I could be wrong of course! 

My main thought here is that now the user can set

Pr, Ra, and Ro_c

and let the code calculate Ek.

aurnou:
On Dec 5, 2023, at 6:23 PM, Loren Matilsky ***@***.***> wrote:﻿
@aurnou, Oh well we don't have to. But my sense is people typically want to set Prandtl to something specific and then play around with the other parameters. It seems unlikely people would want Ra, Ra*, and Ek to all be something specific (thus needing a weird Prandtl number).
I could be wrong of course!
My main thought here is that now the user can set
Pr, Ra, and Ro_c
and let the code calculate Ek.That makes sense to me. Maybe set the default Pr to 1. But it should be a variable, if at all possible.

illorenzo7:
@aurnou the Prandtl_Number should still be variable. I believe its default is still 1. Note that the other defaults are

Rayleigh_Number = -1
Convective_Rossby_Number = -1

etc. We need them to be -1 so that Fortran knows if the user specified them or not.

aurnou:
🙌On Dec 7, 2023, at 5:09 PM, Loren Matilsky ***@***.***> wrote:﻿
@aurnou the Prandtl_Number should still be variable. I believe its default is still 1. Note that the other defaults are
Rayleigh_Number = -1
Convective_Rossby_Number = -1
etc. We need them to be -1 so that Fortran knows if the user specified them or not.

feathern (approved):
I'm good with merging this.  We can work the Prandtl number into the logic in a future PR.

Issue #491, opened 2023-11-30
nconst=10 and nfunc=14 hard-coded into class equation_coefficients


illorenzo7:
This issue concerns the file "reference_tools.py" in the post_processing folder. It hard-codes number of constants and functions in the PDE set to 10 and 14 (the most common) but doesn't allow the possibility that (currently only the constants) these could be different, if the user specifies scalar diffusions. 

I think this could be handled going forward by assuming all previously generated equation_coefficients files had nconst=10, nfunc=14 and keep the current logic if version=1. Then change the version to 2 in Fortran, and have the new equation_coefficients files also include the nconst and nfunc integers explicitly, directly after the version number is written, and before the "cset" and "fset" are written.

illorenzo7:
This issue was fixed in pull request #495

PR #495, opened 2023-12-06
Fix eqcoeff version number

illorenzo7:
Pull request #489 inadvertently broke the equation_coefficients file, and how it is interpreted by post_processing/rayleigh_diagnostics. I should have updated the version number from 1-->2 since I changed the number of constants from 10 to 11. 

This pull request updates the version number, and also the rayleigh_diagnostics class to read in the equation_coefficients file. I verified it still works on old version=1 equation_coefficients files. 

This pull request additionally addresses issue #491, since in version 2, "nconst" and "nfunc" are explicitly written into equation_coefficients.

### DIFF
--- a/post_processing/reference_tools.py
+++ b/post_processing/reference_tools.py
@@ -91,8 +91,6 @@ class equation_coefficients:
             self.nfunc = numpy.fromfile(fd,dtype='int32', count=1)[0]
         else: # if the version is 1, nconst was 10
             self.nconst = 10
-            print("got here")
-            print("nfunc=", self.nfunc)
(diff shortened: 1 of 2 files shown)


feathern (commented):
Looks good, but I have one small suggestion.  Since for version 1, there is no 'dsdr_scale', shouldn't we delete it from the dictionary c_dict on read-in?  Otherwise, I'm ready to approve.

feathern:
Or, alternatively, we could set c11 to 1  for version 1.  Then it won't rescale dsdr, if it's set.

feathern (approved):
Great -- thanks for the small addition.  Merging.

PR #498, opened 2024-03-21
Added return statement to the plot_azav routine.

feathern:
With this PR, the plot_azav routine now returns the img object.  This allows the user to specify the colorbar properties outside of plot_azav if desired.  Previously, there was no return statement, and so the image object needed by the colorbar routine was inaccessible.


illorenzo7 (approved):
Looks good, thanks @feathern !

PR #499, opened 2024-03-21
Update Shell_Avgs to Enable Time-averaging and concatenation

feathern:
This PR modifies the Shell_Avgs class so that it also accepts a list of multiple Shell_Avgs files.  That data can either be concatenated or time-averaged as described in docstring.  In addition, the class now contains an output method so that the current Shell_Avgs object structure can be saved to a new Shell_Avgs file.

Some notes:
1) Single files can still be read as they always have been.
2) This is a lot of new code.  I've been using this new version of the Shell_Avgs class for a couple of years now.   When doing a review, the most straightforward path is probably to test out the concatenation and time-averaging functionality.

### DIFF
--- a/post_processing/rayleigh_diagnostics.py
+++ b/post_processing/rayleigh_diagnostics.py
@@ -992,99 +992,470 @@ class G_Avgs:
         if (closefile):
             self.fd.close()
    
-
-
 class Shell_Avgs:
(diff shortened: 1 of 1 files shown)


illorenzo7 (approved):
Looks good, thanks Nick!

PR #500, opened 2024-03-21
Updated AZ_Avgs class to allow concatenation and time-averaging.

feathern:
This PR modifies the AZ_Avgs class so that it also accepts a list of multiple AZ_Avgs files. That data can either be concatenated or time-averaged as described in docstring. In addition, the class now contains an output method so that the current AZ_Avgs object structure can be saved to a new AZ_Avgs file.

Some notes:
1) Single files can still be read as they always have been.
2) This is a lot of new code. I've been using this new version of the AZ_Avgs class for a couple of years now. When doing a review, the most straightforward path is probably to test out the concatenation and time-averaging functionality.

### DIFF
--- a/post_processing/rayleigh_diagnostics.py
+++ b/post_processing/rayleigh_diagnostics.py
@@ -1087,7 +1087,7 @@ class Shell_Avgs:
         fd.close()
 
 class AZ_Avgs:
-    """Rayleigh AZ_Avgs Structure
+    """Rayleigh Azimuthally-Averaged Data Structure
     ----------------------------------
     self.niter                                    : number of time steps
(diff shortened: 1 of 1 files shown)


illorenzo7 (approved):
Looks good, thanks @feathern

illorenzo7:
@feathern It looks like the AZ_Avgs routine now prints "the_file" any time the class is instantiated from a binary file. I believe that was added in this pull request. Is this behavior strictly necessary or could I make a pull request to remove it? The only reason I care is that it's messing up some of the output from my personal routines (especially the parallelized ones).

feathern:
Oops.  Yes, please do go ahead and submit a PR with the fix.  This was an oversight.

PR #501, opened 2024-04-25
equation_coefficients class can now read old equation_coefficients files (again)

illorenzo7:
This fixes another inadvertent error in the recent pull request #495 which broke the ability of the Python class equation_coefficients() to read the relevant equation_coefficients files with version 1. The problem was that the routine expected the length of the constants array to be 11, but only 10 constants were read in with version 1. This is fixed by simply increasing the length of the constants array and simultaneously entering "1.0" for the last entry.

### DIFF
--- a/post_processing/reference_tools.py
+++ b/post_processing/reference_tools.py
@@ -104,7 +104,9 @@ class equation_coefficients:
         if self.version == 1:
             self.nconst = 11
             self.version = class_version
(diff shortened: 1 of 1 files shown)


feathern (approved):
Sorry this took so long.  I now see that the failed test was related to Mac OS, which seems to be giving problems all around.  This looks good otherwise, so I'm going to approve and merge.

PR #503, opened 2024-05-09
Update rayleigh_diagnostics.py

illorenzo7:
Remove inadvertent print statement of the file name that gets called via AZ_Avgs class.


feathern (approved):
Thanks for catching this.  Approved.

PR #505, opened 2024-05-25
Set f_14 depending on convention; don't reset c's and f's for diffusion types = 3

illorenzo7:
This pull request lumps in the setting of f_14 with c_11 to coincide with different conventions depending on reference_type. It also "protects" various c's and f's from being reset when the user specifies a diffusion using a custom reference (and therefore, set the c and f directly). 

This change should not alter the functionality of the code at all (and it should not alter the internal "reference" object), but it should change the output of the equation_coefficients file in certain circumstances.

### DIFF
--- a/src/Physics/PDE_Coefficients.F90
+++ b/src/Physics/PDE_Coefficients.F90
@@ -2013,13 +2013,21 @@ Contains
         ra_constants(4) = ref%Lorentz_Coeff
         ra_constants(8) = ref%viscous_amp(1)*ref%temperature(1)/2.0d0
         ra_constants(9) = ref%ohmic_amp(1)*ref%density(1)*ref%temperature(1)
+
+        ! the combination c_11 and f_14 cannot be backed out from the "ref" structure
+        ! (which only has ref%dsdr = c_11*f_14)
+        ! and will depend on the convention of the different reference_types
         Select Case(reference_type)
             Case(1,2)
                 ra_constants(11) = 0.0d0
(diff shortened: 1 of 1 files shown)


feathern:
Sorry that it took me a minute to get to this.  I think there's a (maybe) potential bug in that the user can selected reference_type=2 and also use 'with_custom_reference" to specify a background dsdr.   It seems like the logic added here will unset the function and constant associated with dsdr.

illorenzo7:
@feathern Darn, you're right. In fact, I think it was already a bug, because c_11 was already reset to zero in the case you were mentioning. 

I think the cleanest fix would be to make a new flag (say "with_custom_dsdr") that gets set to True if the user sets dsdr via the "with custom" framework. Then only go through the whole block setting c_11 and f_14 if that flag is true. (Otherwise, both c_11 and f_14 were already set). 

What do you think?

illorenzo7:
@feathern. OK, from some limited testing, I think there were a number of bugs after introducing c_11 (beginning I believe with request #489). The first you already pointed out. The second is that c_11 was never set to 1 if the user didn't set it, remaining zero even after "with_custom_functions = 14" or similar was specified in main_input. Both these bugs should now be fixed. I also added some warning statements if the user modifies reference_types 1 or 2 which physically assume dsdr = 0.

feathern (approved):
Great -- thanks for taking care of the other logical bugs.  Approved.

PR #508, opened 2024-06-01
Fairly certain there's a missing c_7 in ohmic heating

illorenzo7:
Similar to the (formerly) missing c_5 in the viscous heating.


feathern (approved):
Thanks for catching this.  Approved.

PR #509, opened 2024-06-11
Update Diagnostics_Thermal_Equation.F90

illorenzo7:
There was a bug in the heat flux (1433) calculation because tmp1d was allocated, but never initialized to zero. Often it would be zero and so the code would run fine, but occasionally NaNs were output in 1433. The initialization should fix this issue.


feathern (approved):
This looks good.  Thanks Loren.  I'm going to go ahead and approve since this change won't affect the tests.

PR #512, opened 2024-06-17
Mention FD

gassmoeller:
The finite difference approximation was not mentioned in the documentation. I added it.


feathern (approved):
Looks good -- approved.

PR #515, opened 2024-06-17
Add dependabot file

gassmoeller:
Github actions has an automatic action that checks the Github actions we are using for newer versions and automatically creates pull requests to update these actions to the latest version. One example of this can be seen here: https://github.com/geodynamics/aspect/pull/5922. I think this would be useful for Rayleigh too to make sure we stay up to date and prevent security issues in the CI pipeline. Just adding this configuration file should enable the bot (it will probably create a bunch of PRs right away, but then will only occasionally find something).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+# Set update schedule for GitHub Actions
+
+version: 2
+updates:
+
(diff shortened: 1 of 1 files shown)


feathern (approved):
Interesting  - never heard of this.  I'm interested to see how it works.  Approved.

PR #524, opened 2024-06-17
Fix ci diff file upload

gassmoeller:
This should hopefully fix the conflict between the uploaded results of the different MPI testers. Each MPI tester now uses its own filename for upload. Let's see what the testers say.


gassmoeller:
Ok, this seems to work now. All CI builds except for the conda builds are working, no more conflicting file names.

tukss (approved):

PR #528, opened 2024-06-17
Bug Fix:  Checkpoint file size

feathern:
This fixes a bug in Rayleigh where a checkpoint file can appear to be too large if the file existed and was overwritten in a subsequent run with lower resolution.   This PR ensures that the file is deleted before being written to, preventing this bug.

### DIFF
--- a/src/IO/Parallel_IO.F90
+++ b/src/IO/Parallel_IO.F90
@@ -1558,6 +1558,7 @@ Contains
         Logical, Intent(In), Optional :: clear_existing
         Integer(kind=MPI_OFFSET_KIND), Intent(In), Optional :: disp
         Integer(kind=MPI_OFFSET_KIND) :: hdisp, tdisp, fdisp, bdisp
+        Integer(kind=MPI_OFFSET_KIND) :: zero_file_size = 0
 
         ! A baseline displacement within the file can be passed to this routine.
         ! Can be due to header and/or multiple prior records.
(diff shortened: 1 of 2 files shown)


src/IO/Parallel_IO.F90
```diff
@@ -1573,6 +1574,11 @@ Subroutine Write_Data(self,disp,file_unit,filename)
             If (present(filename)) Then
                 ! The file is not open.  We must create it.
                 If (self%output_rank) Then
+                    If (present(clear_existing)) Then
+                        If (clear_existing) Then
+                            Call MPI_File_delete(filename, MPI_INFO_NULL, ierr)
```
gassmoeller, line 1579: It looks like you are currently doing this unconditionally no matter if the file is present or not. Can MPI_File_delete handle the case if the file is not present? Or is it only called if the file exists and I missed the place where that is checked?

gassmoeller (commented):
The rationale makes sense to me, but I have a question about the implementation, see above.

src/IO/Parallel_IO.F90
gassmoeller: As discussed this might be easier if you call `MPI_FILE_set_size(funit, 0, ierr)` after the call to `MPI_FILE_OPEN`, this would avoid the case where you dont know if the file already exists (because `MPI_FILE_OPEN` just created it).

gassmoeller (approved):
Looks good to go now, thanks for making the change. Feel free to merge at any time when you are comfortable with the tests on Pleiades.

illorenzo7:
OK, this looks good and fixes the bug I was encountering.

PR #532, opened 2024-06-18
removed references to standard c++ libraries

BrandonJLazard:
Rayleigh does not require references to standard c++ libraries in the configure file in order to make directories.


feathern (approved):
Thanks!  Approved.

PR #533, opened 2024-06-18
Equations solved in Rayleigh

tami1000:
Equations for W,Z,P, A and C, along with equations for nonlinear terms.


feathern:
This looks good to me.  We can always polish later, but having the streamfunction equations added to the documentation is great!

feathern (approved):
Approved (see my earlier comments)

PR #534, opened 2024-06-18
Simplify link flags in configure

gassmoeller:
A suggestion for how to simplify the link options in the configure script. This should be carefully tested on systems with:
- Intel compiler + MKL
- GNU compiler without MKL (this worked on my system)
- GNU compiler with MKL through debian apt
- GNU compiler with MKL through conda
- GNU compiler on MacOS with MKL through conda

### DIFF
--- a/configure
+++ b/configure
@@ -849,31 +849,31 @@ if [[ $USE_MKL == "TRUE" ]]
 then
     if [[ $FVERSION == "INTEL" ]]
     then
-        MKL_LIB="-mkl"
-    fi
-    if [[ $FVERSION == "GNU" || $FVERSION == "AOCC" ]]
+        LIB_FLAGS="-mkl"
+    elif [[ $FVERSION == "GNU" || $FVERSION == "AOCC" ]]
     then
+        # Start by specifying the directory, unless installed by apt on debian
+        # in which case it is in default path
+        if [[ $DEBIANMKL == "FALSE" ]]
+        then
+          LIB_FLAGS="-L${RAMKLROOT}/lib"
+        fi
+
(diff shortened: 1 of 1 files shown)


illorenzo7:
I at least tested this on Pleiades (intel + mkl) and it fixed my issue with the "$MKL_LIB" appearing in configure instead of "-mkl". (So it now compiles with this fix).

feathern:
I just tested the GNU compiler with and without MKL (using Openblas through conda).  This worked fine for me.

feathern:
I also just tested the GNU compiler with MKL installed via apt.  No problems there either.

cianwilson:
I tested this on an intel mac (OS Ventura 13.6.7) using:
```
conda config --set channel_priority strict
conda env create -f environment.yml
conda activate radev
conda install mkl
export MKLROOT=$CONDA_PREFIX
./configure --FC=mpifort --CC=gcc -conda-mkl
make
make install
```
which produced a functioning executable (tested on an `input_example`) despite an `ld` warning:
```
ld: warning: object file (cmkdir.o) was built for newer macOS version (13.0) than being linked (11.5)
```

gassmoeller:
So that seems pretty good then.

> `ld: warning: object file (cmkdir.o) was built for newer macOS version (13.0) than being linked (11.5)`

Hmm, I cannot see how my change could have affected that, the flags for Macos should be exactly as before. Can you check if setting `STD_FLAGS='-lc++'` in the lines above  resolves the warning? Then it would be related to #532 and not this one.

feathern (approved):
This looks good.  We just checked and there was no linking against the standard C++ library.  I'm merging.

PR #535, opened 2024-06-18
Modified Configuration File

BrandonJLazard:
Added conditionals so that the user has the correct compiler and libraries configured before the actual configuration is run.

### DIFF
--- a/configure
+++ b/configure
@@ -477,10 +477,13 @@ done
 
 if [ ! -z $RAMKLROOT ]
   then
+    echo "Correct Configuration"
 elif [[ ! -z $FFTWROOT  &&  ! -z $BLASROOT  &&  ! -z $LPROOT ]];
   then
+    echo "Correct Configuration"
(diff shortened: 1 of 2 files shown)


configure
```diff
@@ -838,9 +854,9 @@ echo ""
 # Standard LIBRARIES
 if [[ $OSTYPE == *"darwin"* ]]
 then
-    STD_FLAGS=""
+    STD_FLAGS="-lc++"
 else
-    STD_FLAGS=""
+    STD_FLAGS="-lstdc++"
```
gassmoeller, line 859: Did you add the STD_FLAGS back in on purpose?

gassmoeller (commented):
Makes sense, just one question.

configure
BrandonJLazard: I did not, not sure why those appeared back up. Will change that

gassmoeller:
The testers dont like it: https://github.com/geodynamics/Rayleigh/actions/runs/9586908067/job/26435825639?pr=535
Have you tested that this works for you?

gassmoeller (approved):
I think this is now good to go.

PR #536, opened 2024-06-19
add CMake support

tukss:
This adds support for building Rayleigh using CMake. It is purely optional and does not interfere with the traditional, `configure`-based approach.

You can try it like this:
```sh
cmake -B build
cmake --build build -j
```
The `rayleigh` executable will be placed in `build/src`.

Todo:

- [x] Fix `Run_Param_Header.F`. The Fortran flags are currently missing. The linker flags have a `SHELL:` prefix.
- [x] Document CMake build procedure including how to make `Debug` and `Release` builds.
- [x] Test on more systems, in particular MKL based ones.
- [x] Add CI tests for the CMake build.
- [x] Install the executable in `bin`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,8 @@ project(rayleigh LANGUAGES C Fortran)
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
 
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
 find_package(MPI COMPONENTS C REQUIRED Fortran REQUIRED)
 find_package(LAPACK COMPONENTS Fortran REQUIRED)
 find_package(FFTW REQUIRED)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.28)
+cmake_minimum_required(VERSION 3.13)
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,11 @@
+cmake_minimum_required(VERSION 3.28)
+
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
+
+project(rayleigh LANGUAGES C Fortran)
+
+find_package(MPI COMPONENTS C REQUIRED Fortran REQUIRED)
+find_package(LAPACK COMPONENTS Fortran REQUIRED)
+find_package(FFTW REQUIRED)
+
+add_subdirectory(src)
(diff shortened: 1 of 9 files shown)


gassmoeller:
Oh, I am glad I didnt make the PR ;-). Did you see https://github.com/geodynamics/Rayleigh/pull/281? I am not sure what we can use from there, but it is worth checking. I am happy to test this branch on my system.

tukss:
Thanks for pointing that out. I've checked and I think the new CMake is already more advanced. In particular the handling of `Run_Param_Headers.F` and MPI.

CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+cmake_minimum_required(VERSION 3.28)
```
gassmoeller, line 1: is 3.28 necessary? I am running Ubuntu 22.04 and that ships 3.22 and I would guess there are plenty of older versions still floating around. In ASPECT we only require 3.13 to not exclude people with unmaintained institutional clusters. That is probably not necessary here, but 3.28 is probably quite uncommon.

tukss:
As a bonus we get support for the new LLVM-based Intel `ifx` compiler for free.

CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
...
+
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
+
+project(rayleigh LANGUAGES C Fortran)
```
gassmoeller, line 5: I would put the project command above the list command.

CMakeLists.txt
gassmoeller, line 1: This file should get a license header as well.

gassmoeller (commented):
The system already works well for me. Some small suggestions for the `CMakeLists`.

CMakeLists.txt
tukss: I've just relaxed it to 3.23. I have to check what we need exactly.

tukss: 3.13 seems to work, too. I'll leave it there for now.

CMakeLists.txt
tukss: Changed that.

tukss:
It even works fine on Cray systems with the cray-fftw. The main thing I still want to try is pleiades.

CMakeLists.txt
tukss: I've added license headers to both cmake files. Actually neither the `configure` nor the `Makefile` have a license header right now.

gassmoeller:
I tested the build on the NSF/TACC systems Stampede3 and Frontera, both use Intel compilers and MKL. I used Intel icc 19.1 on Frontera, and Intel 2024 (the new clang based) on Stampede3. Both worked fine, except I needed to load the fftw3 module.

tukss:
I think this is ready now.

When I added the test, I got an error with the macOS tester because it passed `-fallow-argument-mismatch` to the clang compiler there. The problem goes away by switching to openmpi, which supports the F08 interface. I don't recall why we used mpich before but we should talk about that.

tukss:
@feathern We have a changelog entry now. Feel free to merge it at any time.

feathern (approved):
Looks good -- approved.

PR #537, opened 2024-06-20
Allow radial resolution changes for multiple domains

illorenzo7:
This pull request should accomplish two main goals:

1. Allow the user to increase the number of radial points in each Chebyshev sub-domain (even with multiple domains, e.g., ncheby=64, 96 --> ncheby = 96,256) and then restart from a checkpoint with the old (coarse) resolution. Before, the chebyshev coefficients weren't initialized correctly for multiple domains. 

2. Throw an error message and exit if the user tries to lower radial resolution in any subdomain. Before, the code would simply do nothing, likely integrating a bunch of zeros.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - The streamfunction equations solved by Rayleigh have been added to the documentation.   \[Tami Rogers; 6-20-2024; [#533](https://github.com/geodynamics/Rayleigh/pull/533)\]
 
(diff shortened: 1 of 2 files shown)


illorenzo7:
@feathern OK, this should be ready to be reviewed. I made logic so as to not touch the old code block that handles the single-domain case. The only place single-domain behavior will change is to throw an error and exit if the user tries to lower the resolution. 

FYI I tested this by restarting from a checkpoint in (1) a simulation with a single domain and (2) a simulation with three subdomains. In each simulation, I tried lowering, raising, and not modifying the radial resolution and achieved the expected results.

feathern (approved):
Looks good.  I'm glad you contained the new functionality in its own conditional so that the single-domain checkpointing code remains unchanged!  Approved.

PR #546, opened 2024-06-21
Improve and prettify cmake

gassmoeller:
@tukss: This PR improves and fixes a few issues of the cmake build system. It does:

- remove the `MKL_PARALLEL` option of the FindMKL script. This option enables linking to the threaded MKL libraries, which is not wanted for Rayleigh, and it causes some issues with discovering the FFTW libraries in MKL (at least on Stampede3)
- enable native optimizations for the new Intel clang based compiler ifx
- allow to install Rayleigh into user provided directories (default is still `${CMAKE_SOURCE_DIR}/bin`)
- set the default build type to `Debug` and output build type and install prefix during configuration
- prettify the cmake output


CMakeLists.txt
```diff
@@ -21,10 +21,46 @@ cmake_minimum_required(VERSION 3.13)
...
+# Set up CMAKE_BUILD_TYPE to Debug if it is not set by the user.
+if ("${CMAKE_BUILD_TYPE}" STREQUAL "")
+  set(CMAKE_BUILD_TYPE
+      "Release"
+      CACHE STRING
+      "Choose the type of build, options are: Debug, Release."
+      FORCE)
+  # Set the possible values of build type for cmake-gui
+  set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS
+  "Debug" "Release")
```
tukss, line 43: Let's offer all the normal CMake build types. Especially `RelWithDebInfo` can be very useful.

```suggestion
      "Choose the type of build, options are: Debug, Release, RelWithDebInfo, MinSizeRel."
      FORCE)
  # Set the possible values of build type for cmake-gui
  set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS
  "Debug" "Release" "RelWithDebInfo" "MinSizeRel")
```

tukss (changes requested):
Works well for me. I've made one suggestion.

gassmoeller:
As discussed with @tukss and @feathern the default is now the optimized mode and we offer all build modes.

For reference, this is how my cmake configuration now looks like:

Screenshot from 2024-06-21 12-52-02

tukss (approved):
Looks good now.

feathern (approved):
Looks good.  Approved.

PR #548, opened 2024-06-21
Updated Pleiades install instructions.

feathern:
The logic recently changed in the configure file so that the -mkl flag must be passed to configure when using MKL. This is true even when using an Intel compiler with MKLROOT specified in the user's environment. This PR updates the documentation to reflect this change. In addition, it is best practice on Pleiades to load the compiler and mpi module at the same time owing to a bug that occurs on some node types if they are run one after the other. This has nothing to do with Rayleigh, but I've updated the documentation to hopefully help people avoid this bug.


gassmoeller (approved):
Ready to merge once the documentation tester is done. :+1:

PR #549, opened 2024-06-21
Update Stampede2 instructions to Stampede3

gassmoeller:
Stampede2 was decommissioned earlier this year and has been replaced by Stampede3. Instructions are largely the same, but I changed the recommended configure to cmake (as it worked with less path specifications for me).

I also fixed a bug (I think) of the cmake instructions `--build` needs to be followed by the build directory.


feathern (approved):
Looks good.

PR #550, opened 2024-06-21
Remove unused source file and simplify cmake

gassmoeller:
[src/Physics/Linear_Terms_Cart.F90](https://github.com/geodynamics/Rayleigh/compare/main...gassmoeller:Rayleigh:remove_cartesian_leftovers?expand=1#diff-76968326f9b48904d4108921acd8dd1b30e249d1ed317e09684dcafb035499ee) was unused and outdated. In case a Cartesian version if of interest in the future we can get it back from the git history, but no need to carry an uncompiled file around. This also allows to simplify the cmake system a bit.


src/CMakeLists.txt
```diff
@@ -17,16 +17,9 @@
...
-  Diagnostics/*.F90
-  IO/*.F90
-  Math_Layer/*.F90
-  Parallel_Framework/*.F90
-  Physics/*.F90
-  Test_Suite/*.F90
-  IO/cmkdir.c)
-list(REMOVE_ITEM SOURCES ${PROJECT_SOURCE_DIR}/src/Physics/Linear_Terms_Cart.F90)
+file(GLOB_RECURSE SOURCES LIST_DIRECTORIES true
+  *.F90
```
tukss, line 21: This might be a problem if we have a `build` directory around from a previous build with `configure`. It would build the files twice then. That's why I've spelled out the directories explicitly. Of course, we could also filter our the `build` directory after.

tukss (changes requested):
I'm on board with this. I just didn't dare remove the other file. ;-)

We should be careful about remnant `src/build` directories. See my comment.

src/CMakeLists.txt
gassmoeller: Ah, that makes sense. I was wondering why you spelled them out. Let use the safe option then, I reverted that part of the PR.

tukss (approved):
Looks good now.

PR #551, opened 2024-06-21
Added jupyter notebook for reading in checkpoint files

BrandonJLazard:
Added a jupyter notebook in ~/post_processing that uses the new checkpoint_read function to read in checkpoint files, transform them and plot the results


feathern (approved):
Thanks for this.  This is going to be super helpful for a lot of people!
-Nick

PR #552, opened 2024-06-21
check the  documentation build in the conda action

tukss:
At the moment the build just continues on failed documentation builds in the Conda action. That's not very useful and we might as well skip the documentation build. We should check it though to make sure the docs also build on macOS.

This PR makes the action fail on error and splits the documentation build into HTML (`make doc`) and PDF (`make docpdf`) outputs.


gassmoeller (approved):
So this simply disables the pdf documentation build on the conda tester?

I think this is fine since the readthedocs build still builds the pdf.

PR #553, opened 2024-06-21
Add a symlink to rayleigh_spectral_input.py from post_processing.

cianwilson:


gassmoeller (approved):
Sometimes the simplest solution is the best :smile:

PR #554, opened 2024-06-21
Added documentation for the pseudo-incompressible approximation

BWHindman:
Adds the equation set solved by the pseudo-incompressible approximation to the Rayleigh documentation. This information appears in the Under Development Section.


feathern (approved):
Looks good.  Approving now since the documentation test has passed.

PR #556, opened 2024-06-22
Adding a documentation page for "Utilities"

cianwilson:
This allows for more in-depth, standalone documentation of scripts.

Currently this just includes rayleigh_spectral_input.py but that seems to be in demand.


feathern (approved):
Super useful.  Approved.

PR #558, opened 2024-06-22
Add documentation to the cmake build system

gassmoeller:
This adds the targets `make doc`, `make doc_html`, and `make_pdf` to the cmake build system. `make doc` builds both of the other targets.

The targets depend on all files in `doc/source` so it is only rebuild if something changes in that directory.

The only change to the existing system that I had to make was the modification to the conf.py that makes the exclude path of `doc/build` independent of the directory it is executed from (otherwise I got the same recursive addition of build files we repaired earlier in the week).


tukss (approved):
Works well. Thank you.

PR #559, opened 2024-06-22
Add j2001 hydro test

### LINKED ISSUE #557: More tests for MHD
We should make tests out of the input examples:

- `input_examples/j2011_steady_hydro_minimal`
- `input_examples/j2011_steady_mhd_minimal`

gassmoeller:
Fix #557 by adding the j2011 test cases to the test suite. Let's see how the testers react.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -62,6 +62,14 @@ jobs:
         cd "$GITHUB_WORKSPACE"/tests/c2001_case0_FD_uniform
         mpirun -np 4 ../../bin/rayleigh.dbg
 
+        # J2011 steady hydro test
+        cd "$GITHUB_WORKSPACE"/tests/j2011_steady_hydro_minimal
+        mpirun -np 4 ../../bin/rayleigh.dbg
+
+        # J2011 steady mhd test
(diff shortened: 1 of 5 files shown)


gassmoeller:
Oh, I didnt realize these cases included randomness. How are the random numbers in Rayleigh implemented? Are they reproducible? For ASPECT we had to switch to a different random number generator  (namely https://cplusplus.com/reference/random/mt19937/) to generate reproducibly the same random numbers if we start with the same seed.

feathern:
I would suggest removing MHD test for now with we work on building something with a non-random initial condition.  I just ran the hydro test on my machine and it's good.  Also, I did a quick sanity check to make sure that significant changes in the measured variables were occurring after such a short number of timesteps, and they are, so I think the hydro test is great.  I can approve and merge asap once the MHD test is removed (assuming you haven't already come up with a fix).

feathern (approved):
I'm trying to wrap things up for a pre-workshop release, so I'm going to ahead and approve this, followed by a PR of my own that comments out the lines that execute the MHD test.  We can look at using a different random number generation method at this summer's workshop.

PR #561, opened 2024-06-23
Add option to allow `reference_tools` to load some broken `equation_coefficients` files

### LINKED ISSUE #560: Some past checkpoint files had inconsistent `equation_coefficients`
Some unfortunate users (at least @rpvin possibly others) used Rayleigh versions between 4c76a13c5115a7c9e59725700ecbf5a54c790868 (changed from 10 to 11 constants) and 08ec22ccf826d0e6b0afb734c78ddaeda7195342 (changed the version number in the `equation_coefficients` file). This lead to inconsistent equation coefficients files with version `1` but 11 constants. These can't be read by the Python tools and other versions of Rayleigh making it hard to impossible to reuse these checkpoints.

tukss:
This fixes #560 and includes some Python cleanup.

A simple script like this (save as `convert.py`) can be used to convert all the `equation_coefficients` files that were written in the inconsistent state:
```python
#!/usr/bin/env python3

import sys
from reference_tools import equation_coefficients

for f in sys.argv[1:]:
    e = equation_coefficients(file=f, override_nconst=True)
    e.write(f)
 ```
 
 Run it with `find -name equation_coefficients -exec python3 path/to/convert.py {} +`. Back up your files first.

### DIFF
--- a/post_processing/reference_tools.py
+++ b/post_processing/reference_tools.py
@@ -13,8 +13,20 @@ class equation_coefficients:
                  'eta':7, 'd_ln_rho':8, 'd2_ln_rho':9, 'd_ln_T':10, 'd_ln_nu':11, 'd_ln_kappa':12,
                  'd_ln_eta':13, 'ds_dr':14}
 
-    def __init__(self,radius=[], file=None):
-        if (len(radius) != 0):
+    def __init__(self, radius=[], file=None, override_nconst=False):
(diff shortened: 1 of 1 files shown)


gassmoeller (approved):
Seems good to me.

feathern:
Me too.  Sorry that we didn't catch this bug before it snagged someone.  I'll merge.

PR #564, opened 2024-06-25
[wip] Simplified and minimal update to include pseudo-incompressibility

BWHindman:
Added the capability to switch from anelastic to the pseudo-incompressible approximation.  With pseudo-incompressibility turned off, it has run the hydro anelastic benchmark on Pleiades without throwing errors, but it ran incredible slow.  Made as a draft pull to allow other developers to check the implementation.

### DIFF
--- a/src/Physics/Controls.F90
+++ b/src/Physics/Controls.F90
@@ -79,7 +79,7 @@ Module Controls
     Namelist /Physical_Controls_Namelist/ magnetism, nonlinear, rotation, lorentz_forces, &
                 & viscous_heating, ohmic_heating, advect_reference_state, benchmark_mode, &
                 & benchmark_integration_interval, benchmark_report_interval, &
-                & momentum_advection, inertia, n_active_scalars, n_passive_scalars,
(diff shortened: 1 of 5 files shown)


feathern (approved):
I've looked at this very carefully, and I believe this is good to go.  The implementation looks correct to me both in terms of the code changes and the associated math.  I did check the additional terms in the diffusion coefficients, and based on my understanding of what needed to be chain-ruled, they appear correct to me.  I'm approving this.

feathern:
You'll have to set this to a non-draft before I can merge.  If you want to wait until Pleiades is up so we can test the speed once more, that's fine with me.  Everything you described offline sounds like one of your nodes had a memory problem.

BWHindman:
I believe that I hit all the correct buttons to convert this pull request from a draft to a "ready to merge" request.  Should be good to go.

feathern:
Great.  I just ran some tests to be extra sure, and there is no measurable speed difference between your version and the main-branch version of Rayleigh.  I suspect what you saw on the NASA systems was just another symptom of the old hardware.

PR #566, opened 2024-06-26
Bug Fix:  Diagnostics and Second Radial Derivatives when Using Finite-Difference Mode

feathern:
This fixes a bug related to finite-difference mode.  When diagnostic output quantities are requested that require a second radial derivative, that derivative is calculated at output time.  The subroutine that handles the derivative calculations at output time was lacking any logic that would call the appropriate finite-difference derivative routine.  Instead, a Chebyshev derivative routine was called, which caused a crash because the Chebyshev grid object had not been initialized.  This PR fixes this bug by adding appropriate finite-difference logic to Diagnostics_Second_Derivatives.F90.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


feathern:
I forgot to update the dependencies file originally.  I just did, and now the tests pass.

tukss:
@rpvin Maybe that fixes your crashes?

tukss (approved):
Looks good to me. Thanks for fixing that.

PR #569, opened 2024-07-25
Removed unused variables, module and apply suggestions from Ruff

AlexKurek:


post_processing/rayleigh_diagnostics.py
```diff
@@ -2750,8 +2750,8 @@ def TimeAvg_AZAverages(file_list,ofile):
     ntheta = a.ntheta
     nq = a.nq
     tmp = np.zeros((ntheta,nr,nq),dtype='float64')
-    simtime   = np.zeros(1,dtype='float64')
-    iteration = np.zeros(1,dtype='int32')
+    # simtime   = np.zeros(1,dtype='float64')
+    # iteration = np.zeros(1,dtype='int32')
```
tukss, line 2754: These are just a copy & paste remnant from above. We should delete them.

```suggestion
```

post_processing/rayleigh_diagnostics.py
```diff
@@ -2700,13 +2700,13 @@ def Compile_GlobalAverages(file_list,ofile):
...
-    iters = np.zeros(niter_estimate,dtype='int32')
-    ncount = 0 # total number of iterations read so far
-    ind = 0
+    # niter_estimate = a.niter*nfiles
+    # nq = a.nq
+    # combined = np.zeros((niter_estimate,a.nq),dtype='float64')
+    # time = np.zeros(niter_estimate,dtype='float64')
+    # iters = np.zeros(niter_estimate,dtype='int32')
+    # ncount = 0 # total number of iterations read so far
+    # ind = 0
```
tukss, line 2709: We don't use these temporary buffers because we write to the file on the fly. So these can go.

```suggestion
```

post_processing/rayleigh_diagnostics.py
```diff
@@ -2808,8 +2808,8 @@ def TimeAvg_ShellAverages(file_list,ofile):
         tmp = np.zeros((nr,nq),dtype='float64')
     else:
         tmp = np.zeros((nr,4,nq),dtype='float64')        
-    simtime   = np.zeros(1,dtype='float64')
-    iteration = np.zeros(1,dtype='int32')
+    # simtime   = np.zeros(1,dtype='float64')
+    # iteration = np.zeros(1,dtype='int32')
```
tukss, line 2812: These can go. Again just copy & paste.
```suggestion
```

post_processing/rayleigh_diagnostics.py
```diff
@@ -950,7 +950,7 @@ def read_data(self,qcodes = []):
         if (len(qcodes) == 0):
             self.vals[:,:] = fdata['fdata']['vals'][0,:,:]
         else:
-            nqfile = self.nq        # number of quantity codes in the file
+            # nqfile = self.nq        # number of quantity codes in the file
```
tukss, line 953: We just use `self.nq` here. Let's keep just the comment.
```suggestion
            # number of quantity codes in the file
```

tukss (changes requested):
Thank you for the cleanup. I've reviewed the changes and suggested removing some parts entirely instead of just copying. For others I think keeping the commented variables is better.

The cleanup if the `if`s is particularly appreciated. Thanks!

tukss (approved):
Thanks. Looks good to me now.

AlexKurek:
> The cleanup if the `if`s is particularly appreciated. Thanks!

This is easy if you use a linter, such as Ruff or Pylance. In VS Code you can even use both at the same time as plugins.

Issue #570, opened 2024-08-05
Output error when trying to write out averages (FD)


rpvin:
I am getting index access issues as shown below, when I try to write out averages (Shell, Global and Azimuthal). I have traced the error to vals(self%phi_local(p),r,t), where self%phi_local has zeros and the indices of vals start at 1. Not sure why this is happening now, after I updated to the latest version of Rayleigh. 

forrtl: severe (408): fort: (3): Subscript #1 of the array VALS has value 0 which is less than the lower bound of 1
Image              PC                Routine            Line        Source             
libifcoremt.so.5   00001555553E0F42  for_emit_diagnost     Unknown  Unknown
rayleigh.dbg       00000000004A4482  parallel_io_mp_ca         896  Parallel_IO.F90
rayleigh.dbg       00000000007B1B29  spherical_io_mp_s         730  Spherical_IO.F90
rayleigh.dbg       00000000007AA89F  spherical_io_mp_a         582  Spherical_IO.F90
rayleigh.dbg       00000000009A8868  diagnostics_veloc          43  Diagnostics_Velocity_Field.F90
rayleigh.dbg       00000000010651DE  diagnostics_inter         207  Diagnostics_Interface.F90
rayleigh.dbg       00000000011BF1B5  sphere_physical_s         164  Sphere_Physical_Space.F90
rayleigh.dbg       000000000123F4AA  sphere_driver_mp_         181  Sphere_Driver.F90
rayleigh.dbg       00000000012CBA70  MAIN__                     66  Main.F90
rayleigh.dbg       0000000000402DC2  Unknown               Unknown  Unknown
libc-2.28.so       00001555470557E5  __libc_start_main     Unknown  Unknown
rayleigh.dbg       0000000000402CCE  Unknown               Unknown  Unknown

Also, to get to this error, I had to comment out the following in PDE_coefficients.F90, which was giving a different error, as it was a new addition that wasn't required for the anelastic FD implementation.

!Integrate dsdr to obtain a self-consistent entropy profile. Set s=0 at the upper surface.
ref%entropy(1) = 0.0d0
geofac = (radius(1)**3 - radius(N_R)**3)/3.0d0
Do i = 2, N_R
    ref%entropy(i) = ref%entropy(i-1) - geofac*ref%dsdr(i)*radial_integral_weights(i)/Radius(i)**2
Enddo
ref%exp_entropy = exp(ref%entropy/pressure_specific_heat)
[...]

PR #571, opened 2024-08-13
Added finite-difference support to log_deriv subroutine in PDE_Coefficients.F90.

feathern:
This PR should fix the finite-difference part of issue #570 reported by @rpvin .  Please try this out to see if initialization now works properly for custom reference states with FD turned on.


feathern:
Somehow I made this change in my main branch.   I'm going to close this PR, fix that error, and the reopen a PR.

PR #572, opened 2024-08-15
Added finite-difference support to the log_deriv subroutine.

feathern:
This PR should fix the finite-difference part of issue #570 reported by @rpvin . Please try this out when you get a chance to see if initialization now works properly for custom reference states with FD turned on.


tukss:
There was a missing if guard around the deallocate. I've pushed that directly to your branch.

feathern:
Good catch.  I just moved the deallocate statement into the main Chebyshev "if" block to save a couple of lines of code + keep related things together.

tukss:
That looks good. I'm just running some tests here and will approve this later today if all goes well.

tukss:
It seems to work fine with gfortran but I'm getting a floating-point overflow after the first time step with the Intel compiler (the new LLVM based one). I'm investigating where this comes from.

tukss:
OK. That seems to have been some bad interaction of Cray libsci as the LAPACK provider and the `ifx` compiler. Switching to MKL instead fixes the problem.

tukss (approved):
Works in my test cases.

feathern:
Great -- thanks for checking that.

PR #573, opened 2024-08-15
Checkpointing interval can now be specified in minutes.

feathern:
This PR adds the checkpoint_minutes variable to the temporal_controls namelist.  Similarly to the quicksave_minutes variable, this can be used as an alternative to specifying the number of timesteps between adjacent checkpoints.

I've also updated the documentation to reflect this change.

I'm tagging @jorafb on this as a potential reviewer.


jorafb:
I believe this works!

jorafb (approved):
This works ok!

PR #574, opened 2024-08-19
Consistency Check for Indexed Outputs (e.g., Shell_Slices, Meridional_Slices)

feathern:
This PR implements some consistency/sanity checks when Rayleigh interprets the values of variables such as shellslice_levels, meridional_indices etc.  Previously, Rayleigh would accept values that fell outside of the grid bounds, which could lead to crashes.  With this addition, indicial values that fall outside of the grid bounds are filtered out.   If unallowed values are specified for the normalized indices (e.g., shell_slice_levels_nrm), those should also be filtered out by these modifications to the code logic.

This addresses the cause of the 2nd crash reported in issue #570.

### DIFF
--- a/src/IO/Spherical_IO.F90
+++ b/src/IO/Spherical_IO.F90
@@ -1721,7 +1721,7 @@ Contains
                 jmin =  indices_inout(i)+1
                 
                 ! Only expand the subrange if the values of jmin and jmax
-                ! are consistent.
+                ! are consistent and if they fall within the grid bounds.
                 If ((jmin .ge. 1) .and. (jmax .ge. jmin)) Then
                     Do j = jmin,jmax
                         If (j .le. indmax) Then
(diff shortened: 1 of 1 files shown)


tukss (changes requested):
Thank you for that. I have some small stylistic changes. Feel free to take them or not.

I have tested that it caps subranges as expected. It doesn't actually do anything when we specify single indices outside the valid range, correct? Like setting `meridional_indices = 180` in a simulation with `ntheta = 16`.

feathern:
Thanks for reviewing this.  That's correct, any indices specified outside the allowed ranges are simply ignored.  I can probably add a warning message in a future PR if you want.   I'm fine with all of your suggested changes and just pushed an updated version.  Also, the call to that routine that was commented out was from an earlier version -- there is no longer intended to be a routine named enforce_bounds.  Thanks for catching that.

tukss (approved):
Looks good.

Issue #325, opened 2021-09-27
Output goes to weird places when logfiles are specified because of Write(6,*)


illorenzo7:
@feathern 

In testing multirun mode, I am noticing that although most output from Rayleigh is handled through 

stdout%print()

a fair amount is handled through

Write(6, *)

Normally, this will all go to the same place (terminal screen) so if someone just redirects the screen output, say,

mpiexec 2048 ./rayleigh.opt > logfile

it will be fine. However, if a logfile is specified (which we must do in multirun mode), 

stdout % print 

will go to the logfile, and

write(6, *)

will go to the "screen" which is likely the parent directory. 

Unless I am missing something, it seems like Write(6, *) should never be used in the code, only stdout%print ?

Doing a grep looks like these write(6,*) statements appear 49 times. Happy to make a pull request fixing this (if I haven't missed anything) but it will change quite a few files. Can also wait until version 1.0

illorenzo7:
Fixed by PR #575

PR #575, opened 2024-08-20
Removal of Write(6,*)

feathern:
This PR removes all calls to write(6,\*) in the code that can be encountered during the course of a normal simulation.   Instead, stdout%print is used, which is redirected to a log file, if desired.    This change prevents a situation where some error message could be written out to standard out, whereas others would appear in the log file.   Note that several references to write(6,\*) have been retained.  These are not encountered during normal usage of the code and can still be helpful for debugging.  In addition, many references to write(6,\*) which were commented out or unnecessary have been removed.  This PR fixes issue #325 .


tukss (approved):
That looks good to me. Thanks for cleaning this up!

PR #579, opened 2024-12-05
Add c_10 to code 1434, i.e. vol_heating should = c_10*f_6

illorenzo7:


feathern (approved):
Good catch -- and sorry that I completely missed this PR from so long ago!  Approved.

Issue #423, opened 2022-12-14
Warnings and errors


illorenzo7:
I don't think the code is consistent in its usage of warnings and errors. I believe a "warning" should occur when the code notices something might be up, and prints a useful message, but keeps running. If there is an "error", the code should exit. Currently lots of "errors" don't lead to exits. It looks like most of these occur in the (relatively new) files:

PDE_Coefficients.F90
and
Generic_Input.F90

I'm not sure of best way forward: Either we make the errors that don't lead to exits warnings, or we force the things listed as errors to exit. It might be worth going through the cases in the above files one by one. I plan to do this for PDE_Coefficients after addressing the custom reference stuff: #419 and #420. 

I'm flagging this as a "bug" as well as "errors and messages" because the code isn't running as it should in some cases. That could be a little harsh though...

illorenzo7:
@feathern and @tukss (two people who I feel might have something to add): 

I just checked the source code and I have an update to this issue: For the warnings, no warning resulted in an exit, so I think the warnings are all functioning as intended! For the errors though, there are still multiple errors that don't result in an exit, or do so in different ways. I'm not sure if these are actually supposed to be warnings, or if they are errors that SHOULD result in an exit. 

Specifically: 

src/Physics/Generic_Input.F90, line 186: error but no exit

src/Physics/ProblemSize.F90, line 327: error but no exit

src/Physics/Input.F90
lines 324, 333 for example: There are a few GOTO statements and weird indentation. This results in a "multi-run error"  but no exit. Not sure what the code does here. 
lines 197,198: I think the code exits but does not call pfi%exit(1) as normal, but perhaps that is fine/intended (it does MPI_Finalize(ierr) then Exit(input_error_code)
The indentation was weird in this one and I am less familiar with GOTO usage so perhaps all is good in this file, I don't know. 

src/Physics/PDE_Coefficients.F90: No "error" results in an exit. I will try to fix this in part of my pull requests for this file. 

src/Physics/Checkpointing.F90 (not sure if this already changed with PR #617):
line 661: pfi%exit() (no error code)
line 786: ierr as argument to MPI_BCAST. I assume this does what it's supposed to but I'm not knowledgeable enough to know. 

src/IO/General_IO.F90, line 62: error but no exit

src/IO/Parallel_IO.F90: lines 1706-1731 checks if there is no error, but there is no "else" part of the block, so I don't think this results in an exit

[...]

PR #586, opened 2026-04-07
Draft: Extending custom reference states to active and passive scalar fields.

cianwilson:
This pull request extends custom references states to active and passive scalar fields by adding a new custom reference state file type.  One of these files can be set for active and passive scalar fields and each file contains the reference states for all active or passive scalar fields respectively.

These file names can be set using `chi_a_custom_reference_file` and `chi_p_custom_reference_file` for active and passive scalar fields respectively.  Other new namelist parameters are:
```
    override_chi_a_constant, override_chi_p_constant,
    chi_a_constants, chi_p_constants,
    with_custom_chi_a_constants, with_custom_chi_p_constants,
    with_custom_chi_a_functions, with_custom_chi_p_functions
```
which do the equivalent of their reference state parameters but for active and passive scalar fields.

Using these files it is currently possible to set reference states for:
```
    constants = ['dchirefdr_scale', 'kappa_chi_scale', 'source_chi_scale', 'buoy_chi_scale']
    functions = ['dchirefdr', 'kappa_chi', 'source_chi', 'd_ln_kappa_chi', 'buoy_chi']
```
Note that the buoyancy constant and function will be ignored for passive scalar fields.

Within Rayleigh this also required dchiref and the source terms to be activated (they were previously turned off).

Also fixed some typos and deprecation warnings.

A test has been added in `tests/chi_custom_reference` that provides example input files.  The test runs a case using the existing implementation in temperature and checks that the new implementation for scalar fields produces the same answer.

Documentation has been updated `doc/source/User_Guide/arbitrary_scalar_fields.txt`.


pre_processing/rayleigh_spectral_input.py
```diff
@@ -758,7 +756,7 @@ def parse_mode(modein):
     #--------------------------------------------------------------------------------
     #  )+                       end of grouping
     # - from http://rick.measham.id.au/paste/explain.pl
-    r = re.compile(r'(?:[^,; '+os.linesep+'([{]|\([^)]*\)|\[[^]]*\]|\{[^}]*\})+')
+    r = re.compile(r'(?:[^,; '+os.linesep+'([{]|\\([^)]*\\)|\\[[^]]*\\]|\\{[^}]*\\})+')
```
cianwilson, line 759: This is just a fix for a warning I came across while developing everything else here.

pre_processing/rayleigh_spectral_input.py
```diff
@@ -64,18 +64,20 @@ def dealias_g2m(g):
   if m_max < 0: m_max = 0
   return m_max
 
-def radial_extents(rmin=None, rmax=None, aspect_ratio=None, shell_depth=None):
+def radial_extents(rmin=None, rmax=None, aspect_ratio=None, shell_depth=None, throw=True):
```
cianwilson, line 67: This modification allows this function to be called regardless of whether our function depends on `radius` or not and therefore makes `rmax` and `rmin` available more generally.  It will still throw an error if it can't make the conversion and the conversion is needed (but it won't throw if the conversion is still not needed).

Though unrelated to the PR specifically it is a necessary change for the test.

src/Physics/PDE_Coefficients.F90
```diff
@@ -1191,6 +1271,146 @@ Subroutine Augment_Reference()
...
+            if (active) Then
+                Call stdout%print('Active scalar field reference state will be augmented.')
+                Call stdout%print('Only source, buoyancy, or background dchidr may be modified.')
+            else
+                Call stdout%print('Passive scalar field reference state will be augmented.')
+                Call stdout%print('Only source or background dchidr may be modified.')
+            endif
+            Call stdout%print('Source requires both d_3 and g_3 to be set.')
+            if (active) Call stdout%print('Buoyancy requires both d_4 and g_5 to be set.')
+            Call stdout%print('Reference dchidr requires g_1 to be set (sets d_11 to 1 if unspecified).')
```
illorenzo7: I think this should be d_1 not d_11.

cianwilson: Should be fixed now, thanks!

src/Physics/PDE_Coefficients.F90
```diff
@@ -1581,6 +1822,274 @@ Subroutine Read_Custom_Reference_File(filename)
...
+
+        If (pi_integer .eq. 314) Then
+
+            ! Read in constants and their 'set' flags
+            Read(15) eqversion
+            Read(15) dummy ! n_chi_constants
+            Read(15) dummy ! n_chi_functions
+            Read(15) n_scalars
+            if (n_scalars .gt. n_l_scalars) then
+                Call stdout%print('ERROR: number of scalars in the custom reference file')
```
illorenzo7: I think since the code continues to try to read the file, it should be a "WARNING" (this is part of the larger warning/error issue of #423 so not unique to this part of the code). Also perhaps the print message should say what the code does in this case? It looks like it ignores n_l_scalars from main_input and just reads whatever is in the custom file...

src/Physics/PDE_Coefficients.F90
```diff
@@ -1581,6 +1822,274 @@ Subroutine Read_Custom_Reference_File(filename)
...
+                Call stdout%print('exceeds the number of '//Adjustl(chi_l_type_str)//' scalars')
+                Call stdout%print('specified in main_input.')
+                Call stdout%print('Check your scalar custom reference file and main_input settings.')
+            end if
+            Read(15) ((cset(i, j), i = 1, n_chi_constants), j = 1, n_scalars)
+            Read(15) ((fset(i, j), i = 1, n_chi_functions), j = 1, n_scalars)
+            Read(15) ((input_constants(i, j), i = 1, n_chi_constants), j = 1, n_scalars)
+            
+            ! Cset(i) is 1 if a constant(i) was set; it is 0 otherwise.
```
illorenzo7: make this set(i,j) in comments...

cianwilson: Fixed, thanks!

src/Physics/PDE_Coefficients.F90
cianwilson: I've modified the ERROR -> WARNING and stated what will happen if you carry on.

I've also tried changing the logic underneath so that `n_scalars` and `n_l_scalars` are used appropriately.  This is untested however as all the tests have `nscalars == n_l_scalars`.

illorenzo7 (commented):
I just checked all this over the last few days and it looks good to me! Merging.

illorenzo7 (approved):
I just checked all this over the last few days and it looks good to me! Merging.

cianwilson:
Thanks @illorenzo7 !

PR #591, opened 2026-05-05
[DRAFT] updated sphinxcontrib-bibtex version

feathern:
Some of the failures we are seeing in our tests look as though they are due to an incompatibility with the sphinxcontrib-bibtex package and one of its dependencies, pybtex.  We probably either need to roll the sphinxcontrib package forward, or specify the version of pybtex we want in the yml file.  This draft PR updates the sphinxcontrib version.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -15,7 +15,7 @@ dependencies:
   - jupyter=1.0
   - scipy=1.10
   - sphinx=4.0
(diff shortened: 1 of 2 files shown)


feathern:
That seemed to do the trick.  I'm going to go ahead and merge this so I can see if we can clean up the outstanding dependabot PRs.

PR #595, opened 2026-05-08
[DRAFT]  Testing out how the Jones 2011 hydro (only) benchmark works with the testers. 

feathern:
I think this failed before as well, but the MHD benchmark, which has random initial conditions, has been removed as a test for the time being.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -62,9 +62,9 @@ jobs:
         cd "$GITHUB_WORKSPACE"/tests/c2001_case0_FD_uniform
         mpirun -np 4 ../../bin/rayleigh.dbg
 
-        #J2011 steady hydro test
-         cd "$GITHUB_WORKSPACE"/tests/j2011_steady_hydro_minimal
-         mpirun -np 4 ../../bin/rayleigh.dbg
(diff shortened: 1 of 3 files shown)


feathern:
Hi there @tukss @illorenzo7 @cianwilson @gassmoeller .  

Two things here.  (1)  I have the Jones 2011 hydro benchmark working as a test.  Earlier, we were seeing some failures in some containers, but not others.  I think  this may have been due to the high-resolution (for Github), which may have been causing memory issues.  I bumped the resolution down by a factor of 2x and recreated the benchmark file.  This seems to have fixed that problem.

(2) I also played around with the .yml structure for the tests.  I noticed that all of the tests were being wrapped together under the same "tab," for lack of a better word, in Github.  I broke these out, but I'm wondering if there's a way to create sub tabs.  If you look at the checks on this draft PR, you can see that now we have dedicated arrows for the various runs (e.g., finite-difference test, generic input test, etc.).  

If you have any thoughts on how we might get sub-tabs/sub-arrows, let me know!  Also, thoughts on what I named the different tests are useful.   We may want to use a different naming scheme. 

Anyway, just thinking out loud -- please comment if/when you have time.  No rush on this.

PR #598, opened 2026-05-13
Updating Ubuntu version used by Readthedocs

feathern:
Ubuntu 20.04 is being deprecated by Readthedocs.  This PR updates to 24.04.


feathern:
Looks like this worked fine.  Since this is a small change and the deprecation takes place in two weeks, I'm going to go ahead and approve/merge.

PR #599, opened 2026-05-13
Updates zenodo tag with correct doi for v1.3.0

ljhwang:
Updated doi for new release.


feathern:
Oh, good catch.  Thanks!

feathern (approved):

PR #608, opened 2026-07-12
do not build TACC image anymore

### LINKED ISSUE #602: TACC docker images build is broken
Currently our CI pipeline fails at the sanity checks in the `configure` script. I think that is due to changes made to the TACC base container, specifically because MKL is not included under `/opt` anymore. We can fix that with a bit of effort but the broader question is if we want to keep supporting this particular container or maybe even extend it to other systems. I personally don't have access to TACC but could work on supporting Athena and Perlmutter.

tukss:
We would need some changes to make it build with the TACC base image. Remove this for now as we do not seem to have users for these containers at the moment.

Also removes the respective documentation to avoid confusion. If we want to resurrect that feature, we can just take it out of the git history.

Fixes #602.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -73,30 +73,3 @@ jobs:
           tags: |
             ghcr.io/${{env.REPO_NAME_LC}}:${{github.ref_name}}
             ${{env.REPO_NAME_LC}}:${{github.ref_name}}
-
-
-      - name: Build and push TACC image for main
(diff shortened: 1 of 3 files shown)


tukss:
Unfortunately we'll only see if this works after merging because that's when the container builds run.

feathern (approved):
This looks reasonable to me.  Let's see what the testers do...

PR #609, opened 2026-07-13
Initalizing compressible model variables in controls and fields

BrandonJLazard:
Added physical control variables for the compressible model as well as initialized the eq variables in fields in a slightly different manner to account for the difference which variables are being solved for. 

Added an additional PDE coefficient for the compressible momentum eq.


feathern (approved):
Looks good.  Approved.

PR #611, opened 2026-07-13
Compressible Model Implementation into Rayleigh

BrandonJLazard:
Added alternative subroutines for the compressible model


BrandonJLazard:
Per Nick's request pushed a committ that has the remainder of the files needed for the compressible model to run

src/Physics/Controls.F90
```diff
@@ -124,8 +128,8 @@ Module Controls
     Integer :: num_quicksaves = 3              ! Number of quick-save checkpoints to write before rolling back to #1
     Real*8  :: quicksave_minutes = -1.0d0      ! Time in minutes between quick saves (overrides quicksave interval)
     Real*8  :: checkpoint_minutes = -1.0d0     ! Time in minutes between checkpoints (overrides quicksave interval)
-
-    Real*8  :: cflmax = 0.6d0, cflmin = 0.4d0  ! Limits for the cfl condition
+    
+    Real*8  :: cflmax = 0.6d0, cflmin = 0.1d0  ! Limits for the cfl condition
```
feathern: The value of cflmin can be set through a namelist, so let's not change the default (which would affect non-compressible models as well).  Can you revert it to 0.4d0?

src/Physics/Sphere_Hybrid_Space.F90
```diff
@@ -63,37 +63,69 @@ End Subroutine Hybrid_Init
...
         Do mp = my_mp%min, my_mp%max
             SBUFFA(l_max,:,:,:) = 0.0d0
         Enddo
 
-
         ! Allocate two work arrays
         Call Allocate_rlm_Field(ftemp1)
         Call Allocate_rlm_Field(ftemp2)
 
+        !Come back to this
```
feathern: Remove this comment (I think it's out of date).

feathern (approved):
Massive effort.  Approved!

PR #612, opened 2026-07-13
Bugfix in rayleigh_diagnostics (Shell_Slices)

BWHindman:
When reading a single iteration, single quantity code, and single radius (using the keyword slice_spec), the code did not skip to the  end of the record before reading the time and the iteration number.  Hence, those two quantities were reading random stuff.  Added two lines to reposition the read position in the file.


feathern (approved):
Good catch.  Approved.

PR #614, opened 2026-07-13
npcol and nprow were mixed-up in the tutorial

martin11235:
npcol and nprow were mixed-up in Run Rayleigh tutorial page.


tukss (approved):
Thanks for fixing this.

PR #618, opened 2026-07-15
Rewrite of Shell_Spectra in rayleigh_diagnostics.py to allow partial reading of the file

BWHindman:
Rewrote the Shell_Spectra routine in rayleigh_diagnostics.py to allow the user to specify that a subset of the data in the file be read. The new version allows the user to choose which iterations, which quantity codes, which radii, and which set of ells and m should be returned.


illorenzo7 (approved):
Brad and I have checked the routines fairly extensively over the last few days and everything appears to functioning normally/as intended! I'm approving this for merging but want to see if @feathern has any comments before doing the merge.

feathern:
This looks good to me, but I have one comment.   In the previous version, each record was read at once.  In the new version, each shell contained in a record is read individually.  For situations where the full spectrum data is read, this means that the number of reads has increased by a factor of nq*nr.    It might be good to run a few quick timing tests on the old and new versions.  If there's a significant difference, maybe we could retain the old read method in a different logical branch that's only activated when the full file is requested?

illorenzo7:
OK @BWHindman @feathern I just checked the timing of this stuff. I read 5 shell_spectra with problem sizes

ntheta = [2048, 1024, 768, 512, 384]

(also nr = 9, nq = 8, niter = 1). 

I am attaching a plot of the timing results between the main branch Shell_Spectra class and the new class from the pull request. Long story short, there actually is some significant slowdown (about 20%) for certain problem sizes. Specifically corresponding to the above ntheta, the slowdowns are 

array([23.9831828 , 20.39500945, -0.45722965, 23.49546726,  5.51846248]) (%)

I ran each test 3 times and took the average of the timings. 

I'm not sure why certain problem sizes give different results but didn't look too far into it. 

For now, I think we should do what Nick said and make the new class called 

Shell_Spectra_Partial

or something and then keep the old Shell_Spectra routine as is, while we figure out the source of the slowdowns (I feel like they could be avoided somehow). 

The good news is, I ran Brad's new routine choosing just one radius, 1 quantity, 20 l-vals, and 10 m-vals and the speed up was enormous! For the above problem sizes, the new routine read things 

array([588.14485864, 309.00820322, 241.7455058 , 130.82819021, 80.30761453])

times faster! So that's good :) 
comparison

feathern (approved):
Great -- thanks for making those changes.  Approved.

PR #619, opened 2026-07-15
Refactoring the calculation of viscous forces

cianwilson:
Viscous forces are now pre-computed before other diagnostics and stored in vforce_buffer.

This shouldn't affect the behavior of the code and is intended to prepare for the calculation of the curl of the viscous force.

Added a `vforce_diagnostics` test that is a pure regression test with comparison data generated from the main branch.


feathern (approved):
Looks good to me.

Issue #620, opened 2026-07-15
Fortran runtime error with point probes when using debugging


cianwilson:
I'm getting a fortran runtime error when adding point probes to a simulation and running with debugging enabled.

To reproduce add:
```
&output_namelist
point_probe_values = 1228,1229,1230,1231,1232,1233,1234,1235,1236
point_probe_r_nrm = 0.0, -1.0
point_probe_theta_nrm = 0.0, -1.0
point_probe_phi_nrm = 0.20, -0.30, 0.7, -0.8
point_probe_frequency = 100
point_probe_nrec = 1
/
```
to the `main_input` file for `tests/c2001_case0` and run with `mpiexec -np 4 ../../bin/rayleigh.dbg`.

This produces:
```
At line 453 of file Parallel_IO.F90
Fortran runtime error: Index '32' of dimension 1 of array 'tmp' above upper bound of 31

--------------------------------------------------------------------------
Primary job  terminated normally, but 1 process returned
a non-zero exit code. Per user-direction, the job has been aborted.
--------------------------------------------------------------------------
#0  0x7ffff8147ae0 in ???
#1  0x7ffff8148659 in ???
#2  0x7ffff8148c56 in ???
#3  0x5555555db7ac in __parallel_io_MOD_load_balance_io
        at /workspaces/rayleigh.test/src/build/Parallel_IO.F90:453
#4  0x5555555e3374 in __parallel_io_MOD_initialize_io_buffer
        at /workspaces/rayleigh.test/src/build/Parallel_IO.F90:278
#5  0x55555575d071 in __spherical_io_MOD_initialize_diagnostic_info
        at /workspaces/rayleigh.test/src/build/Spherical_IO.F90:1125
#6  0x555555770aeb in __spherical_io_MOD_initialize_spherical_io
        at /workspaces/rayleigh.test/src/build/Spherical_IO.F90:382
#7  0x555555b05f98 in __diagnostics_interface_MOD_initialize_diagnostics
        at /workspaces/rayleigh.test/src/build/Diagnostics_Interface.F90:329
#8  0x555555c516f5 in main_initialization
        at /workspaces/rayleigh.test/src/build/Main.F90:95
#9  0x555555c511c0 in MAIN__
        at /workspaces/rayleigh.test/src/build/Main.F90:64
#10  0x555555c519ac in main
        at /workspaces/rayleigh.test/src/build/Main.F90:31
--------------------------------------------------------------------------
[...]

PR #622, opened 2026-07-16
Point_Probes Can Now Sample all Points in Theta

feathern:
A temporary array used when initializing the io_buffer object was allocated with 1 less element than intended.  This prevented Point_Probes from sampling all theta values local to a particular process.  For example., if a process was responsible for 32 theta points, only 31 of those points could be sampled without producing a seg fault.   This PR increases the size that of array by 1.


cianwilson (approved):
I've tested this on my failing test and it now runs and resolves issue #620  .  Thanks @feathern

PR #624, opened 2026-07-16
Bug Fix for SPH_Modes Class in rayleigh_diagnostics.py

feathern:
The SPH_Modes initialization routine in rayleigh_diagnostics.py was treating the data as though it were striped in a different form that what is actually output by Rayleigh.

In Rayleigh, within a record, the real portion of the entire record (all modes, radii, quantity codes) is written out, followed by the imaginary part.  The SPH_Modes class was treating the data as though each quantity code, rather than each record, was written real part then imaginary part.

This PR fixes the bug by reording the quantity-code and real/imaginary loops used to read in the data.


cianwilson (approved):
Looks good!

PR #625, opened 2026-07-17
Add curl momentum diagnostics

Ayesha714:
I have added curls of advection, buoyancy, magnetic, coriolis, and pressure forces and tested against my own cases.


cianwilson:
Hi @Ayesha714 ,

Hope your trip back is going well.

I've just pushed a first draft (completely untested) implementation of the derivatives of the viscous force and an example implementation of the curl components `curl_viscous_force_r`, `curl_viscous_force_theta` and `curl_viscous_force_phi`.

Any chance you could try them out when you get back to work?  I'm expecting that they will crash and I will have to debug them a bit more so please just let me know what fails.

You can also see the proposed updated documentation of the quantity codes here:
https://rayleigh-documentation--625.org.readthedocs.build/en/625/doc/source/diagnostic_codes/curl_momentum_equation.html
There are a few issues there to look into too.

Thank you!
Cian

Ayesha714:
One more clarification: I have not uncommented the lines related to the curl terms yet because these routines are not currently implemented in `Diagnostics_Curl_Momentum.F90`. I have only made the changes needed for the current diagnostics structure. Once the correct implementation/procedure is clear, I can add and enable the remaining terms.

cianwilson:
Hi @Ayesha714 ,

The diagnostics we've added should be available using the quantity codes you implemented on this branch. So if you compile your branch of the code and run it you should be able to add, for example, a global average of 1327, which should be the r component of the curl of the viscous force term, to the main input file.

Obviously, as you pointed out, those codes may yet change as we update the branch with the missing diagnostics.

Hope you got back home safely eventually!
Cheers,
Cian

PR #626, opened 2026-07-17
Added missing constants into the turb KE codes

illorenzo7:
There were some missing constants in the turbulent Kinetic energy codes. Hopefully @BWHindman or @feathern can quickly convince themselves these are correct!


feathern (approved):
Looks correct -- good catch.

PR #627, opened 2026-07-17
Added missing c_11s to thermal energy equation codes in doc

illorenzo7:
Also took the opportunity to put the v_r fields after all the constants and functions, as is standard! @feathern or @cianwilson should be able to review this quickly.


cianwilson (approved):
Checked that this matches what's in the code and documentation.  Looks good!

PR #630, opened 2026-07-17
Nvfortran compatibility

tukss:
This allows us to build with `nvfortran` and `nvc` from NVHPC. As there is no quad precision support in `nvfortran`, I've added a fallback to double in the Legendre Polynomials. We should probably change that in the long run.

Options:
- Compile that one file with `gfortran`. Con: Is annoying to implement in the build system and might cause interface issues.
- Dump the coefficients once and read them in.
- Call out to `libquad` or similar.


feathern (approved):
Looks good!

PR #631, opened 2026-07-17
Added a temporal FFT transform class to spectral_utils.py

BWHindman:
Added a temporal FFT transform class to spectral_utils.py.  Handles the transforms (forward and back), windowing, spectral filtering, and interpolation to a regular grid.

### DIFF
--- a/post_processing/spectral_utils.py
+++ b/post_processing/spectral_utils.py
@@ -4,14 +4,20 @@ Interface to the Rayleigh grid and spectral transforms, including:
     + Fourier grid/transforms/derivatives
     + Legendre grid/transforms/derivatives
     + Chebyshev grid/transforms/derivatives
+    + Temporal transforms with windowing and grid interpolation (Added 07/2026)
 
 R. Orvedahl May 24, 2021
 """
 from __future__ import print_function
 import numpy as np
+from scipy.fft import fft, fftfreq, ifft
+from scipy.interpolate import make_interp_spline, interp1d
 from scipy.linalg.blas import dgemm as DGEMM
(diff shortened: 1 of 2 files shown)


post_processing/spectral_utils.py
```diff
@@ -2717,4 +2723,405 @@ def transform(self, data_in, input_config, output_config, th_l_axis=0, phi_m_axi
             raise ValueError(e)
 
         return data_out
+        
+
```
tukss, line 2727:
```suggestion


```
Let's get right of trailing spaces.

post_processing/spectral_utils.py
```diff
@@ -2717,4 +2723,405 @@ def transform(self, data_in, input_config, output_config, th_l_axis=0, phi_m_axi
...
+      try:
+        w = np.asarray(self.window)
+
+        if len(w) != n_t:
+          raise ValueError(f"Window size (={len(w)}) does not match data size (={n_t}).")
+        shape = [1] * self.data.ndim
+        shape[self.axis] = n_t
+        windowed_data = self.data * w.reshape(shape)
+      except Exception as e:
+        raise TypeError(f"Window must be one of: None, string, or array-like. Error: {e}")
```
tukss: I would avoid general exception catching like that. Is tahre a reason we can't just capture the exception we expect? Or just let it pass up to the user.

The problem with capturing every exception it might even capture `InterruptException` (when you press Ctrl-C). So the user would get a confusing error message when they just interrupt the calculation.

tukss (changes requested):
Looks good in general. Have a look at my comments.

feathern (approved):
Great addition!

PR #632, opened 2026-07-18
Kappa, nu, and eta are now taken from the custom reference file.

martin11235:
By default, kappa_type, nu_type, and eta_type is 1, an option that ignores the custom reference state. Since the examples/custom_reference_states/MESA-input-1Msun-ZAMS.ipynb notebook is setting radial profile of these variables, type 3 needs to be added in &Transport_Namelist.


tukss (approved):
LGTM. Thanks

PR #633, opened 2026-07-21
A fix for a  bug in Compute_Second_Derivatives when using finite differences.

cianwilson:
Spotted a possible bug in Compute_Second_Derivatives.  Committing this so I remember it but need to test.


feathern (approved):
Hi Cian,

This was absolutely a bug, and your fix should do the trick, so I don't think we need to test anything more here.  As it was written, the p1a buffer was incorrectly being set to zero when in finite-difference mode.  Your fix takes care of that perfectly.  Good catch!

Approved.

-Nick

feathern:
(though I need you to unset it from 'draft' in order to merge)

cianwilson:
Thanks for taking a look @feathern !  'Draft' should be unset now.

PR #635, opened 2026-07-22
Bug squashing for the compressible version

kyle-augustson:
Fixed some missing geometric factors in the compressible velocity diffusion terms and found a mislabeled velocity derivative in the compressible velocity advection terms.

This permits the j2011 hydro benchmark mode 6 to run up to 127k iterations (at max_time_step = 1e-2, at larger timestep it dies at the same simulation time ~ 1270 seconds)  before the timestep reaches 1e-13 (probably a resolution issue). This is about the sound crossing time for the model, so there probably is an unresolved sound wave bouncing around that causes it to peter out.

### DIFF
--- a/src/Physics/Sphere_Physical_Space.F90
+++ b/src/Physics/Sphere_Physical_Space.F90
@@ -256,7 +256,7 @@ Contains
 
                 Call Velocity_Advection()
                 !Write(6, *) "VELCOITY ADVECT", Maxval(RHSP)
-                !Call Velocity_Diffusion()
+                Call Velocity_Diffusion()
                 !Write(6, *) "VELOCITY DIFFUSE", Maxval(RHSP)
                 Call Pressure_Force()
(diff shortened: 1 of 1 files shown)


feathern (approved):
Good catch on this.  I agree with the revised math as well.

BWHindman:
Sounds like the rogue sound wave was launched by your initial conditions.  I have several suggestions:
1) Start with very smooth initial conditions.  That should eliminate the short wavelength waves that might be causing you difficulty.
2) Initialize with a divergenceless velocity field instead of a temperature fluctuation. Don't start with sound waves, let those develop as the convection develops.
3) If you have many scale heights in the domain, there is potentially a large factor of density in all of the eigenfunctions. A small amplitude wave launched deep in the atmosphere can grow many orders of magnitude as it travels upwards.  You might consider either limiting the initial fluctuations either in magnitude or in spatial distribution.  Only perturb in the upper part of the domain.

PR #638, opened 2026-07-23
Bug Fix: IO Buffer Used Before Initialization

feathern:
This PR addresses a bug that was causing the IO Buffer's print method to be called before the buffer was initialized.  This bug seemed to only arise when reference_type=4 was used along with benchmarking mode.   What's odd is that it was causing a crash in the custom-reference-state test, but that crash wasn't causing the tester to throw an error.

It would be good to review/approve quickly if you have a minute @tukss @illorenzo7 @cianwilson

### DIFF
--- a/src/Physics/Benchmarking.F90
+++ b/src/Physics/Benchmarking.F90
@@ -378,9 +378,6 @@ Contains
             reference_type = 4
             custom_reference_file = file_remember
             heating_type = type_remember
-            If (global_rank .eq. 0) Then
-                Call stdout%print('Reference Type 4 set for Benchmark Mode.')
-            Endif
         Endif
         
         If (with_custom_reference) Then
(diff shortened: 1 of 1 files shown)


feathern:
Just a quick note, you can see the crash clearly in the custom-reference testing section of my draft PR #637 .  It looks like I introduced this bug two years ago, but I'm still unclear on why the tester is passing even with the crash.

src/Physics/Benchmarking.F90
```diff
@@ -586,7 +583,10 @@ Subroutine Initialize_Benchmarking
                 Call stdout%print(" ")
                 Call stdout%print(" -- Benchmarking Mode is Activated.")
                 Call stdout%print(" -- Selected Benchmark :  "//trim(benchmark_name))
-                Call stdout%print(" ")
+                If (reference_type .eq. 4) Then
+                    Call stdout%print(' -- Reference Type 4 set for Benchmark Mode.')
+                Endif
+                Call stdout%print(" ")
```
tukss: This line has trailing spaces in the modified version.
```suggestion
                Call stdout%print(" ")
```

tukss (approved):
Makes sense. Not sure why it didn't crash in the test either. Have a look at the stylistic comment.

feathern:
Great.  I accepted the change and will merge once the tests clear again.  Thanks for the quick review.

illorenzo7:
Thanks @tukss and @feathern! Sorry you were too quick for me

PR #640, opened 2026-08-31
Fixing tailing latex doc strings in thermal_equation_codes.F.

cianwilson:
When running doc/build_menu_pdf.py on a different branch I noticed it generated some unrelated changes in thermal_equation.rst.

I checked the code and believe the changes in thermal_equation_codes.F were made in error at some point (but not synced to the documentation).


feathern (approved):
Thanks for catching this.   Looks good.
-Nick

PR #641, opened 2026-09-03
Ordering of compute_BandCurlB and Hybrid_Output_Final had switched

cianwilson:
This PR returns the order of compute_BandCurlB and Hybrid_Output_Final to what it was until recently.  Was there a reason it was changed?

After merging main into a branch I started seeing something that wasn't B in my output (presumably it was C).  This reversion fixes that and gets me B back again.  Once that other branch is finished there will be a test for this but I wanted to get a fix in for this ASAP in case anyone interprets the output as B in the meantime.

### DIFF
--- a/src/Physics/Sphere_Hybrid_Space.F90
+++ b/src/Physics/Sphere_Hybrid_Space.F90
@@ -123,9 +123,9 @@ Contains
            end do
         Endif 
         
-        If (output_iteration) Call Hybrid_Output_Final()
-
         If (magnetism) Call compute_BandCurlB()
+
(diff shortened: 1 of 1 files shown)


cianwilson:
This appears to have happened over two commits:
https://github.com/geodynamics/Rayleigh/commit/3d3b3080bbb7a84ceedcbec94d848b6a73f324dd
https://github.com/geodynamics/Rayleigh/commit/9ee03ea2305d4bf4ba0518fd8cb53b705c3eedcb

feathern (approved):
Good catch!  At least this only crept in during the hackathon, so it hasn't had time to cause many problems yet.

PR #642, opened 2026-09-03
nell was left uninitialized in some diagnostics

cianwilson:
Looking for another bug a valgrind run pointed out that some diagnostic arrays were being allocated based on an uninitialized variable that turned out to be nell for diagnostics that don't depend on it.

This is a hopefully simple fix for that valgrind warning.


feathern (approved):
Good catch.  I wonder if this is related to some isolated instances we've seen here were a single MPI rank mysteriously runs out of memory on startup, causing a crash.  Approved.

PR #644, opened 2026-09-10
Fix scalar field checkpoint restart and spurious active scalar results when magnetism=.true.

cianwilson:
Read_Checkpoint's ab_offset double-counted the scalar field count, leaving the scalar fields unflagged for restart whenever magnetism=.true., so scalar fields silently reset to 0 at start.

Separately, emfphi was aliased to avar+1, which Add_Field now assigns to chiavar(1) whenever n_active_scalars >= 1; Adjust_Emf/Finalize_EMF then overwrote chi_a(1)'s RHS with EMF data every iteration, producing spurious results whenever magnetism was on.

Also fixes an unrelated indexing bug in Link_Equations that doesn't seem to affect anything.

Extends tests/chi_scalar to cover magnetism+scalar interaction and restart transitions that toggle magnetism or scalar fields on/off, which would have caught both bugs.


cianwilson:
Looks like I haven't quite fixed whatever was wrong with checkpointing yet (still a jump across the restart):
image
so hold off on merging this just yet.

cianwilson:
Turns out that checkpointing failure was an MPI issue not related to magnetism or active scalar fields.  I've opened a separate issue for it at #645 .

I think this means that this PR can continue to be reviewed.

Thanks!

Issue #645, opened 2026-09-12
Checkpoint failures on stampede3


cianwilson:
Following the instructions at:
https://rayleigh-documentation--625.org.readthedocs.build/en/625/doc/source/User_Guide/getting_started.html?highlight=stampede#tacc-stampede3
and using the default modules on stampede3 (only xalt and cmake have changed version and python has been added from the link above):
```
login2.stampede3(1078)$ module list

Currently Loaded Modules:
  1) intel/24.0   2) impi/21.11   3) autotools/1.4   4) cmake/4.1.1   5) xalt/3.1.1   6) python/3.12.11   7) TACC
```
and the cmake build:
```
FC="${TACC_FC}" CC="${TACC_CC}" cmake -DCMAKE_BUILD_TYPE=Debug -DRAYLEIGH_CPU_OPTIMIZATIONS=native -Bbuild
cmake --build build -t install
```
so that I end up compiling with the intel oneapi 2021 MPI:
```
login2.stampede3(1091)$ grep MPI_Fortran_COMPILER_INCLUDE_DIRS: build/CMakeCache.txt 
MPI_Fortran_COMPILER_INCLUDE_DIRS:STRING=/opt/intel/oneapi/mpi/2021.11/include/mpi;/opt/intel/oneapi/mpi/2021.11/include;/opt/intel/oneapi/mpi/2021.11/include/mpi
```
I am seeing significant non-deterministic failures in writing checkpoints.  These failures manifest themselves as the checkpoint files not containing the correct output but no crash of the checkpointing run and the failure becomes obvious when I restart.

For example in this pure-hydro (but with an active scalar field) comparison with magic:
<img width="567" height="280" alt="Image" src="https://github.com/user-attachments/assets/575dff89-877c-4fe1-b1b1-9a86244955bf" />
where the collapse in kinetic energy in the Rayleigh simulation occurs immediately after a restart from a checkpoint.

Or, in this much shorter MHD simulation:
<img width="640" height="850" alt="Image" src="https://github.com/user-attachments/assets/539ed32d-5f7f-4641-92b4-3c59fc58cb6f" />
which has a pickup from a checkpoint at the dotted vertical line after just 1000 iterations and shows non-physical jumps in entropy (really T) and chi_a_1 immediately after. 

[...]

PR #646, opened 2026-09-13
Possible fix for race condition on MPI IO

cianwilson:
Add an MPI Barrier to prevent processors from racing each other with certain MPI implementations.

This is a possible fix for #645.  Open to more elegant solutions.


feathern (approved):
Looks good -- hope this prevents headaches for others in the future.

